PR #6, opened 2026-04-19
[codex] Implement constraint-aware .mlfp ergonomics

### LINKED ISSUE #14: Implement constraint-aware .mlfp ergonomics
Tracking issue for PR #6: https://github.com/soulomoon/mlf2/pull/6

## Scope

Implement constraint-aware `.mlfp` ergonomics while keeping raw eMLF unchanged.

## Expected work

- Add constrained types for definitions, class methods, and instance declarations.
- Lower constraints to hidden runtime method evidence.
- Extend instance resolution for constrained schemas such as `Eq a => Eq (Option a)`.
- Reject overlapping instance heads by unification.
- Add parameterized `deriving Eq`, including recursive data types such as `List a`.
- Add qualified and aliased imports.
- Expand the explicit Prelude and document the updated language surface.

## Validation target

The implementation should pass the focused `.mlfp` program tests and full `cabal build all && cabal test` suite.

soulomoon:
## Summary

- Add `.mlfp` constrained types for definitions, class methods, and instance declarations, lowering constraints to hidden runtime method evidence while keeping raw eMLF unchanged.
- Extend instance resolution for constrained schemas such as `Eq a => Eq (Option a)`, reject overlapping heads by unification, and add parameterized `deriving Eq` including recursive `List a`.
- Add qualified and aliased imports, expand the explicit Prelude, and document the updated language surface.

## Validation

- `cabal test mlf2-test --test-show-details=direct --test-options='--match "parameterized deriving Eq for recursive List"'`
- `cabal test mlf2-test --test-show-details=direct --test-options='--match "MLF.Program"'` — 88 examples, 0 failures
- `cabal test mlf2-test --test-show-details=direct --test-options='--match "MLF.Program eMLF"'` — 51 examples, 0 failures
- `cabal build all && cabal test` — 1637 examples, 0 failures
- `git diff --check`

Closes #14

### DIFF
--- a/src/MLF/Frontend/Program/Check.hs
+++ b/src/MLF/Frontend/Program/Check.hs
@@ -783,29 +783,44 @@ synthesizeDerivedInstances scope mod0 = concat <$> mapM deriveForData (moduleDat
       fst (splitArrows (snd (splitForalls (P.constructorDeclType ctor))))
 
     validateEqDerivingField dataDecl fieldTy
-      | fieldCoveredByDerivedConstraints dataDecl fieldTy = pure ()
+      | eqTypeSatisfiable dataDecl Set.empty fieldTy = pure ()
+      | otherwise = throwError (ProgramDerivingMissingFieldInstance "Eq" fieldTy)
+
+    eqTypeSatisfiable dataDecl seen fieldTy
+      | fieldCoveredByDerivedConstraints dataDecl fieldTy = True
+      | key `Set.member` seen = False
       | otherwise =
           case resolveInstanceInfoWithSubst (scopeToElaborateScope scope) "Eq" fieldTy of
-            Right _ -> pure ()
-            Left _ -> throwError (ProgramDerivingMissingFieldInstance "Eq" fieldTy)
+            Right (instanceInfo, subst) ->
+              let seen' = Set.insert key seen
+               in all
+                    (eqConstraintSatisfiable dataDecl seen' . applyConstraintSubst subst)
+                    (instanceConstraints instanceInfo)
+            Left _ -> False
+      where
+        key = ("Eq", show fieldTy)
+
+    eqConstraintSatisfiable dataDecl seen constraint
+      | P.constraintClassName constraint == "Eq" =
+          eqTypeSatisfiable dataDecl seen (P.constraintType constraint)
+      | otherwise = False
+
+    applyConstraintSubst subst constraint =
+      constraint
+        { P.constraintType =
+            Map.foldrWithKey substituteTypeVar (P.constraintType constraint) subst
+        }
 
     fieldCoveredByDerivedConstraints dataDecl fieldTy =
       case fieldTy of
         STVar name -> name `elem` P.dataDeclParams dataDecl
-        STBase name -> name == P.dataDeclName dataDecl
-        STCon name _ -> name == P.dataDeclName dataDecl
-        _ -> False
+        _ -> isRecursiveOwnerField dataDecl fieldTy
 
     scopeToElaborateScope scope0 =
       mkElaborateScope (scopeValues scope0) (scopeTypes scope0) (scopeClasses scope0) (scopeInstances scope0)
 
     mkEqInstance dataDecl =
-      let actualHead = case P.dataDeclParams dataDecl of
-            [] -> STBase (P.dataDeclName dataDecl)
-            (param0 : paramsRest) -> STCon (P.dataDeclName dataDecl) (STVar param0 :| map STVar paramsRest)
-          headTy = case P.dataDeclParams dataDecl of
-            [] -> STBase (P.dataDeclName dataDecl)
-            _ -> actualHead
+      let headTy = dataDeclHeadType dataDecl
           left = P.Param "left" (Just headTy)
           right = P.Param "right" (Just headTy)
           selfName = "__derived_eq_" ++ P.dataDeclName dataDecl
@@ -875,10 +890,12 @@ synthesizeDerivedInstances scope mod0 = concat <$> mapM deriveForData (moduleDat
            in P.EApp (P.EApp (P.EVar eqName) (P.EVar l)) (P.EVar r)
 
     isRecursiveOwnerField dataDecl argTy =
-      case argTy of
-        STBase name -> name == P.dataDeclName dataDecl
-        STCon name _ -> name == P.dataDeclName dataDecl
-        _ -> False
+      argTy == dataDeclHeadType dataDecl
+
+    dataDeclHeadType dataDecl =
+      case P.dataDeclParams dataDecl of
+        [] -> STBase (P.dataDeclName dataDecl)
+        param0 : paramsRest -> STCon (P.dataDeclName dataDecl) (STVar param0 :| map STVar paramsRest)
 
 buildInstanceSkeletons :: Scope -> P.Module -> [P.InstanceDecl] -> TcM [InstanceInfo]
 buildInstanceSkeletons scope mod0 derived = do
--- a/src/MLF/Frontend/Program/Check.hs
+++ b/src/MLF/Frontend/Program/Check.hs
@@ -833,6 +833,27 @@ synthesizeDerivedInstances scope mod0 = do
         STVar name -> name `elem` P.dataDeclParams dataDecl
         _ -> isRecursiveOwnerField dataDecl fieldTy
 
+    derivedConstraintParams dataDecl =
+      let params = Set.fromList (P.dataDeclParams dataDecl)
+          fieldTypes =
+            filter
+              (not . isRecursiveOwnerField dataDecl)
+              (concatMap constructorFieldTypes (P.dataDeclConstructors dataDecl))
+          usedParams = Set.intersection params (foldMap freeTypeVars fieldTypes)
+       in [paramName | paramName <- P.dataDeclParams dataDecl, paramName `Set.member` usedParams]
+
+    freeTypeVars ty =
+      case ty of
+        STVar name -> Set.singleton name
+        STArrow dom cod -> freeTypeVars dom `Set.union` freeTypeVars cod
+        STBase {} -> Set.empty
+        STCon _ args -> foldMap freeTypeVars args
+        STForall name mb body ->
+          maybe Set.empty (freeTypeVars . unSrcBound) mb
+            `Set.union` Set.delete name (freeTypeVars body)
+        STMu name body -> Set.delete name (freeTypeVars body)
+        STBottom -> Set.empty
+
     scopeToElaborateScope scope0 =
       mkElaborateScope (scopeValues scope0) (scopeTypes scope0) (scopeClasses scope0) (scopeInstances scope0)
 
@@ -855,7 +876,7 @@ synthesizeDerivedInstances scope mod0 = do
             { P.instanceDeclClass = "Eq",
               P.instanceDeclConstraints =
                 [ P.ClassConstraint "Eq" (STVar paramName)
-                  | paramName <- P.dataDeclParams dataDecl
+                  | paramName <- derivedConstraintParams dataDecl
                 ],
               P.instanceDeclType = headTy,
               P.instanceDeclMethods = [P.MethodDef "eq" methodBody]
--- a/src/MLF/Frontend/Program/Check.hs
+++ b/src/MLF/Frontend/Program/Check.hs
@@ -873,7 +873,14 @@ buildInstanceSkeletons scope mod0 derived = do
   let instances0 = derived ++ explicitInstances mod0
   infos <- mapM toInstanceInfo instances0
   ensureDistinctPlain (\(className0, ty) -> ProgramDuplicateInstance className0 ty) [(instanceClassName info, instanceHeadType info) | info <- infos]
+  case duplicateExistingInstances infos of
+    info : _ -> throwError (ProgramDuplicateInstance (instanceClassName info) (instanceHeadType info))
+    [] -> pure ()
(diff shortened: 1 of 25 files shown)


src/MLF/Frontend/Program/Check.hs
```diff
@@ -481,48 +796,87 @@ synthesizeDerivedInstances scope mod0 = concat <$> mapM deriveForData (moduleDat
...
+        STBase name -> name == P.dataDeclName dataDecl
+        STCon name _ -> name == P.dataDeclName dataDecl
+        _ -> False
 
 buildInstanceSkeletons :: Scope -> P.Module -> [P.InstanceDecl] -> TcM [InstanceInfo]
 buildInstanceSkeletons scope mod0 derived = do
   let instances0 = derived ++ explicitInstances mod0
   infos <- mapM toInstanceInfo instances0
   ensureDistinctPlain (\(className0, ty) -> ProgramDuplicateInstance className0 ty) [(instanceClassName info, instanceHeadType info) | info <- infos]
+  case overlappingInstances infos of
```
soulomoon, line 1221: `buildInstanceSkeletons` only checks overlap inside the current module's `infos`, but `scope0` already contains imported/prior instances. This lets a module import `Eq a => Eq (Box a)` and then define `Eq (Box Bool)` without any error as long as the instance is unused, despite the PR's coherence contract that overlapping heads are rejected by unification. Include `scopeInstances scope` in the overlap check before accepting the local skeletons.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -691,99 +869,165 @@ compileCatchAllOnly scope mbExpected scrutineeExpr alts =
...
+
+    compileCandidates _ [] = throwError (ProgramNonExhaustiveCase [ctorName ctorInfo])
+    compileCandidates topArgs [alt] =
+      compileAltCandidate topArgs alt Nothing
+    compileCandidates topArgs (alt : rest) = do
+      fallback <- compileCandidates topArgs rest
+      compileAltCandidate topArgs alt (Just fallback)
+
+    compileAltCandidate topArgs (P.Alt pattern0 body) mbFallback =
+      case stripPatternAnn pattern0 of
```
soulomoon, line 1219: All `PatAnn` cases strip the annotation and continue as if it was not present, so `case Zero of { (_ : Bool) -> true }` succeeds even though the scrutinee is `Nat`. Since the PR documents pattern annotations as supported source syntax, the checker needs to validate the annotated pattern type against the matched scrutinee/field type instead of ignoring it.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -691,99 +869,165 @@ compileCatchAllOnly scope mbExpected scrutineeExpr alts =
...
 
 ctorOwners :: [P.Alt] -> [String]
 ctorOwners = foldr go []
   where
     go alt acc = case P.altPattern alt of
       P.PatCtor ctorName0 _ -> ctorName0 : acc
+      P.PatAnn inner _ -> go (P.Alt inner (P.altExpr alt)) acc
       _ -> acc
 
+validateOrderedPatterns :: [P.Alt] -> ElaborateM ()
```
soulomoon, line 1365: After recording a flat constructor catch-all such as `Succ _`, the validator only rejects a later branch when that later branch is also a flat catch-all. A later `Succ Zero` passes validation if a final `_` makes the generated fallback exhaustive, so `Succ _; Succ Zero; _` runs with the unreachable branch silently ignored. Branches after a constructor-local catch-all should be rejected regardless of whether the later branch is specific or catch-all.

soulomoon (commented):
I found three correctness issues in the new `.mlfp` ergonomics slice. The two P1s break the stated coherence/typechecking contract, and the P2 lets an unreachable branch pass silently.

src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -612,9 +698,11 @@ resolveDeferredMethods scope deferredMethods = go
             case inferClassArgument (lowerType scope (methodType methodInfo)) (methodParamName methodInfo) argTypes of
               Just ty -> Right ty
               Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
-          instanceInfo <- resolveInstanceInfo scope (methodClassName methodInfo) classArgTy
-          runtimeName <- concreteMethodRuntimeName instanceInfo methodInfo
-          Right (foldl X.EApp (X.EVar runtimeName) args)
+          (instanceInfo, subst) <- resolveInstanceInfoWithSubst scope (methodClassName methodInfo) classArgTy
+          methodValue <- concreteMethodValue instanceInfo methodInfo
+          evidenceArgs <- resolveConstraintEvidenceTerms scope Set.empty (map (applyConstraintSubst subst) (instanceConstraints instanceInfo))
+          let methodHead = instantiateMethodValue scope subst methodValue
```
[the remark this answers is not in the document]
soulomoon: @codex address that feedback

src/MLF/Frontend/Program/Run.hs
```diff
@@ -82,8 +98,17 @@ stripUnusedTopTyAbs term = case term of
   EUnroll body -> EUnroll (stripUnusedTopTyAbs body)
   _ -> term
 
+toValueWithProgram :: CheckedProgram -> ElabTerm -> Value
+toValueWithProgram checked term =
+  case mainSourceType checked >>= \srcTy -> decodeSourceValue checked srcTy term of
+    Just value -> value
+    Nothing ->
+      case decodeAnyData checked term of
+        Just value -> value
```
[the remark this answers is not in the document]
soulomoon: @codex address that feedback

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -682,108 +867,223 @@ localIdentityScrutinee scope expr =
...
+compileCatchAllOnly :: ElaborateScope -> Maybe SrcType -> Maybe SrcType -> SurfaceExpr -> [P.Alt] -> ElaborateM SurfaceExpr
+compileCatchAllOnly scope mbExpected mbScrutineeTy scrutineeExpr alts =
   case alts of
     [P.Alt P.PatWildcard body] -> compileExpr scope mbExpected body
     [P.Alt (P.PatVar name) body] -> do
       runtimeName <- freshRuntimeName name
       scope' <- extendLocalLowered scope name runtimeName =<< freshTypeName
       bodyExpr <- compileExpr scope' mbExpected body
       pure (surfaceLet runtimeName scrutineeExpr bodyExpr)
+    [P.Alt (P.PatAnn inner _) body] -> compileCatchAllOnly scope mbExpected mbScrutineeTy scrutineeExpr [P.Alt inner body]
```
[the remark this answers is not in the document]
soulomoon: @codex address that feedback

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

soulomoon:
@codex review

src/MLF/Frontend/Program/Check.hs
```diff
@@ -267,19 +300,432 @@ buildImportScope priorExports imports0 = foldM go emptyScope imports0
...
+qualifiedInstanceVariants alias exports unqualifiedTypeNames unqualifiedClassNames instanceInfo =
+  distinctInstanceHeads
+    [ variant
+      | variant <- fullQualifiedVariants ++ aliasHeadVariants,
+        not (sameInstanceHead instanceInfo variant)
+    ]
+  where
+    qualifiedInstance = qualifyInstance alias exports instanceInfo
+    aliasHeadNeeded =
+      instanceClassName instanceInfo `Set.member` unqualifiedClassNames
```
soulomoon: This guard still compares only the raw class name when deciding whether to synthesize an alias-head instance. If `Core` exports only `Token(..)` but defines its own private `class Eq`/`instance Eq Token`, and `Main` imports an unrelated `Classes exposing (Eq, eq)` plus `Core as C`, `unqualifiedClassNames` contains `Eq` and this emits an `Eq C.Token` candidate backed by Core's hidden class instance. Then `eq C.Token C.Token` can resolve against the private Core instance instead of failing for the imported `Classes.Eq`. The variant needs to be guarded by class identity/origin, not just the unqualified class name.

soulomoon (commented):
I found one remaining instance scoping issue.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -196,11 +230,12 @@ checkModule priorModules mod0 = do
   valueScope <- liftEither =<< pure (addValues (scopeValues importScope) localValues)
   typeScope <- liftEither =<< pure (addTypes (scopeTypes importScope) localData)
   classScope <- liftEither =<< pure (addClasses (scopeClasses importScope) localClasses)
-  let scope0 = Scope valueScope typeScope classScope (scopeInstances importScope ++ priorInstances)
+  let scope0 = Scope valueScope typeScope classScope (scopeInstances importScope ++ priorInstances ++ qualifiedPriorInstances)
```
soulomoon: This still adds every earlier module's `priorInstances` to the current module, regardless of what the module imports. That means a private/unimported instance from an earlier module can satisfy a later module's same-named local class/type by raw name alone; for example, `Core` can define a private `class Eq` plus `instance Eq Token`, and a later `Main` that does not import `Core` but defines its own `class Eq` and `data Token` can have `eq Token Token` resolved using `Core`'s instance instead of failing with `ProgramNoMatchingInstance`. The instance scope needs to be derived from instances made visible by the current imports (with class identity and canonical type identity preserved), not from all prior modules wholesale.

soulomoon (commented):
Found one remaining scoping issue.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -267,19 +301,521 @@ buildImportScope priorExports imports0 = foldM go emptyScope imports0
...
+
+visibleInstancesForImports ::
+  Map P.ModuleName ModuleExports ->
+  Map P.ModuleName (Map String DataInfo) ->
+  [InstanceInfo] ->
+  Set.Set ClassIdentity ->
+  [P.Import] ->
+  [InstanceInfo]
+visibleInstancesForImports priorExports priorData priorInstances unqualifiedClassIdentities =
+  distinctInstanceHeads . concatMap instancesForImport
```
soulomoon, line 510: This new cross-import dedupe still keys equivalent instances by raw `instanceClassName` plus `instanceHeadType`, so it can discard a distinct instance for another class that happens to have the same unqualified name. For example, if `A` exports only method `z` for its private `class Eq` with `instance Eq Int`, and `B` exports only method `a` for its private `class Eq` with `instance Eq Int`, then `Main` importing `A exposing (z)` and `B exposing (a)` loses B's instance here and `a 1 1` fails with `ProgramNoMatchingInstance`. Please include the class identity (`instanceClassModule`, unqualified class name) in the dedupe key, or avoid deduping across different class identities.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -851,14 +1518,92 @@ isRecursiveResultType ty =
...
+
+    classInfoIdentity classInfo =
+      (classModule classInfo, unqualifiedName (className classInfo))
+
+    instanceInfoClassIdentity info =
+      (instanceClassModule info, unqualifiedName (instanceClassName info))
+
+    overloadedMethodClassIdentity =
+      methodValueClassIdentity
+        =<< find visibleMethodForClass (Map.elems (esValues scope))
```
soulomoon: This fallback picks the first visible method with the same raw `methodClassName`, not the actual overloaded method currently being resolved. With two exported methods from separate hidden same-named classes, resolving `eqB 1 1` can pick A's class identity, select A's instance, and then fail with `ProgramUnknownMethod "eqB"` or use the wrong evidence. The resolver needs the class identity carried by the specific `MethodInfo`/`ValueInfo` that created the deferred call instead of scanning `esValues` by raw class name.

soulomoon (commented):
I found two remaining class-identity edge cases in the instance visibility changes.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -538,41 +1240,182 @@ buildInstanceSkeletons scope mod0 derived = do
...
+          instanceClassName left == instanceClassName right,
+          instanceHeadType left /= instanceHeadType right,
+          instanceHeadsOverlap (instanceHeadType left) (instanceHeadType right)
+      ]
+
+    duplicateExistingInstances infos =
+      [ local
+        | local <- infos,
+          existing <- scopeInstances scope,
+          instanceClassName local == instanceClassName existing,
```
soulomoon: This local-vs-imported duplicate check still compares only the surface class name, so it treats imported instances for hidden same-named classes as duplicates of a local class instance. For example, after `import A exposing (z)` and `import B exposing (a)` where both modules export methods from private `class Eq` instances over `Int`, a `Main` module that defines its own local `class Eq` plus `instance Eq Int` is rejected with `ProgramDuplicateInstance "Eq" (STBase "Int")` even though the three instances belong to distinct class identities. The overlap check below has the same raw-name guard; these comparisons need to use `instanceClassIdentity` like `sameInstanceHead` now does.

soulomoon (commented):
I found one remaining class-identity issue in the local-vs-imported instance checks.

Issue #28, opened 2026-04-22
Migrate .mlfp checking and elaboration to consume resolved symbols


soulomoon:
Parent: #12

Scope:
Replace downstream qualified/unqualified string branching in `.mlfp` checking, elaboration, and runtime metadata with resolved identities from the early resolver. Keep source display/rendering and runtime name generation explicit.

Acceptance criteria:
- `MLF.Frontend.Program.Check` no longer infers semantic identity from string qualification in normal value/type/class/constructor/method paths.
- Instance visibility, duplicate/overlap checks, deriving, constructor-result validation, and case checking compare semantic identities rather than surface spellings.
- Existing qualified/aliased import coverage in `ProgramSpec` continues to pass.
- Regression coverage shows mixed qualified/unqualified spelling does not create duplicate value, constructor, class, type, method, or instance identities.
- Architecture or language-reference documentation records the resolved-symbol boundary where behavior or ownership changes.

Dependencies/blockers:
- Depends on both preceding #12 sub-issues: the resolved-symbol model and the resolver pass.
- Coordinate with active LSP work if shared analysis results need to expose resolved symbols or resolver diagnostics.

Sibling compatibility:
This sub-issue owns downstream consumer migration only. It must not redesign the resolver model, change parser syntax, implement #9 method disambiguation, or introduce #8 higher-kinded type behavior; tests/docs should validate this migration without duplicating sibling resolver/model scope.

soulomoon:
Implemented by merged PR #33.

PR #33, opened 2026-04-22
Implement #28

soulomoon:
Implements #28.

# Implementation Plan for soulomoon/mlf2#28

1. Establish a resolver-backed checked-module context.
   - Change `checkProgram`/`checkModules` so the `ResolvedProgram` produced by `resolveProgram` is threaded into module checking instead of being stored only after the old checker completes.
   - For each module, build lookup maps from `ResolvedModule` scopes/references keyed by surface spelling and by `SymbolIdentity`.
   - Keep parser syntax, resolver semantics, runtime name generation, and source display names unchanged.

2. Attach semantic identities to downstream metadata.
   - Extend `ValueInfo`, `DataInfo`, `ConstructorInfo`, `ClassInfo`, `MethodInfo`, and `InstanceInfo` with stable identity fields or small accessor helpers backed by `SymbolIdentity`/`SymbolOwnerIdentity`.
   - Populate local metadata from existing constructors and imported metadata from resolver symbols rather than deriving semantic ownership from qualified strings.
   - Keep visible map keys as surface spellings only for lookup and diagnostics; all semantic comparisons should use identities.

3. Replace qualified-copy import semantics in `MLF.Frontend.Program.Check`.
   - Rework `buildImportScope`, `applyImportItem`, `qualifyModuleExports`, and instance-visibility helpers so alias imports create visible spellings pointing at the same semantic metadata, not cloned declarations with qualified semantic names.
   - Preserve explicit source display names for diagnostics/rendering and preserve `valueRuntimeName`, `ctorRuntimeName`, and method runtime names.
   - Deduplicate visible/imported values, constructors, types, classes, methods, and instances by semantic identity where mixed qualified/unqualified spellings refer to the same declaration, while still rejecting truly conflicting visible names.

4. Move type/class/constructor comparisons to identity-aware canonicalization.
   - Update constructor-result validation, deriving validation, instance duplicate/overlap checks, and `canonicalInstanceHead` so type heads compare by `DataInfo` identity rather than `STBase`/`STCon` spelling.
   - Update class and method checks to compare `classInfoSymbolIdentity`, `methodInfoSymbolIdentity`, and `instanceInfoClassSymbolIdentity` instead of `unqualifiedName` string stripping.
   - Keep overlap unification behavior otherwise unchanged.

5. Update elaboration and finalization consumers.
   - In `MLF.Frontend.Program.Elaborate`, make `canonicalSourceType`, `lowerTypeRaw` callers, constructor owner lookup, pattern/case matching, overloaded method resolution, and evidence lookup use metadata identities.
   - Ensure `constructorNameMatches`, `sameConstructorInfo`, `sameDataInfo`, and `resolveConstructorDataInfo` compare semantic identities so `Zero` and `C.Zero` can coexist as spellings for one constructor in the same case.
   - In `Finalize`, audit deferred constructor/case/method resolution for remaining spelling-derived semantic checks and convert those to identity-backed metadata where needed.

6. Add regression coverage in `ProgramSpec` and focused symbol tests.
   - Keep existing qualified/aliased import matrix cases passing.
   - Add mixed spelling regressions proving no duplicate semantic identities are created for values, constructors, classes, types, methods, and instances when imported both unqualified and through an alias.
   - Add behavior tests for constructor-result validation, deriving, case checking, and instance duplicate/overlap checks across mixed qualified/unqualified spellings.
   - Add one resolved-program assertion that mixed spellings produce distinct `SymbolSpelling` values but one `SymbolIdentity`.

7. Document the boundary.
   - Update `docs/architecture.md` with the resolved-symbol boundary: resolver owns semantic identity; checker/elaborator use surface spelling only for lookup, diagnostics, source rendering, and runtime-name construction.
   - Update the note in `MLF.Frontend.Program.Types` so it no longer says the model is unused/pending after the migration lands.

8. Validate and publish in implementation mode.
   - Run a narrow targeted test first, e.g. `cabal test mlf2-test --test-show-details=direct --test-options='--match "MLF.Program"'`.
   - Then run the required behavior-changing gate: `cabal build all && cabal test`.
   - Before committing, inspect `git status --short` and relevant diffs; stage only implementation, tests, and docs for this issue.

### DIFF
--- a/src/MLF/Frontend/Program/Elaborate.hs
+++ b/src/MLF/Frontend/Program/Elaborate.hs
@@ -496,7 +496,7 @@ explicitExprAnnotation expr =
 
 compileValueApp :: ElaborateScope -> Maybe SrcType -> ValueInfo -> [P.Expr] -> ElaborateM SurfaceExpr
 compileValueApp scope mbExpected ConstructorValue {valueCtorInfo = ctorInfo} args = do
-  let expectedArgTys = constructorExpectedArgTypes scope ctorInfo args
+  let expectedArgTys = constructorExpectedArgTypes scope ctorInfo mbExpected args
   argSurfaces <-
     zipWithM compileConstructorArg expectedArgTys args
   placeholder <- deferConstructorCall scope ctorInfo (length args) mbExpected
@@ -569,10 +569,15 @@ isPartialOverloadedMethodApp scope expr =
           not (null args) && length args < methodFullArity methodInfo
     _ -> False
 
-constructorExpectedArgTypes :: ElaborateScope -> ConstructorInfo -> [P.Expr] -> [SrcType]
-constructorExpectedArgTypes scope ctorInfo args =
-  reverse (snd (foldl step (Map.empty, []) (zip (ctorArgs ctorInfo) args)))
+constructorExpectedArgTypes :: ElaborateScope -> ConstructorInfo -> Maybe SrcType -> [P.Expr] -> [SrcType]
+constructorExpectedArgTypes scope ctorInfo mbExpected args =
+  reverse (snd (foldl step (initialSubst, []) (zip (ctorArgs ctorInfo) args)))
   where
+    initialSubst =
+      case mbExpected >>= matchTypesInScope scope Map.empty (constructorOccurrenceType ctorInfo (length args)) of
+        Just subst -> subst
+        Nothing -> Map.empty
+
     step (subst, acc) (templateTy, arg) =
       let expectedTy = specializeSrcType subst templateTy
           subst' =
--- a/src/MLF/Frontend/Program/Elaborate.hs
+++ b/src/MLF/Frontend/Program/Elaborate.hs
@@ -506,6 +506,7 @@ compileValueApp scope mbExpected ConstructorValue {valueCtorInfo = ctorInfo} arg
       case inferKnownExprType scope arg of
         Just knownTy
           | Set.null (freeTypeVarsSrcType knownTy) ->
+              ensureSourceTypeCompatible scope expectedTy knownTy >>
               compileExpr scope (Just knownTy) arg
           | otherwise ->
               compileExpr scope (Just expectedTy) arg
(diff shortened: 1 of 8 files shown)


src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -1572,10 +1589,10 @@ resolveInstanceInfoWithIdentity scope className0 expectedClassIdentity headTy =
         Nothing -> False
 
     instanceInfoClassIdentity info =
-      (instanceClassModule info, unqualifiedName (instanceClassName info))
+      instanceInfoClassSymbolIdentity info
 
     matchInstanceHead info =
-      case matchTypes Map.empty (instanceHeadType info) headTy of
+      case matchTypesInScope scope Map.empty (instanceHeadType info) headTy of
```
soulomoon, line 1699: With this identity-aware head match, the same imported instance can now match through both the unqualified and qualified variants. `equivalentInstanceMatch` still compares `instanceConstraints` after only lowering the constraint type, so `Eq a` and `C.Eq a` are treated as different constraints even when their `ClassInfo` identities are the same. A constrained instance imported both ways, for example `instance Eq a => Eq (Box a)`, will produce two non-equivalent matches and resolve as `ProgramNoMatchingInstance`. Please canonicalize constraint class identities here as well, or dedupe the qualified instance variants semantically.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -502,6 +506,7 @@ compileValueApp scope mbExpected ConstructorValue {valueCtorInfo = ctorInfo} arg
       case inferKnownExprType scope arg of
         Just knownTy
           | Set.null (freeTypeVarsSrcType knownTy) ->
+              ensureSourceTypeCompatible scope expectedTy knownTy >>
```
soulomoon: This guard still misses generic constructor arguments when the constructor result type supplies the type variable. `expectedTy` is the unspecialized constructor-field template here, so for a constructor like `Use : A.Box a -> Use a` in a context expecting `Use Bool`, an argument known as `B.Box Bool` compares against `A.Box a`; the lowered types are not equal yet because of `a`, so `sourceTypesNeedNominalRejection` returns false. Later `deferConstructorCall` uses the result expectation to bind `a ~ Bool`, and the eMLF check can accept `B.Box Bool` for `A.Box Bool` because the encodings are the same. Please specialize the constructor argument types from `mbExpected` before this compatibility check, or validate again after `deferredConstructorInitialSubst` is known.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -502,6 +506,7 @@ compileValueApp scope mbExpected ConstructorValue {valueCtorInfo = ctorInfo} arg
...
               compileExpr scope (Just knownTy) arg
           | otherwise ->
```
soulomoon: The result-based `initialSubst` fixes concrete constructor applications, but polymorphic constructor values still bypass the nominal check through this branch. For example, if `B.Empty : B.Box a` is passed to `Use : A.Box a -> Use a` in a `Use Bool` context, `inferKnownExprType` returns `B.Box a`, so `Set.null (freeTypeVarsSrcType knownTy)` is false and we compile it against `A.Box Bool` without calling `ensureSourceTypeCompatible`. Because the deferred constructor later recovers from same-shaped Church encodings, alias/map ordering can let that `B.Box Bool` be accepted for `A.Box Bool`. Please run the compatibility check after specializing `knownTy` with the expected type, or add an explicit nominal validation for this free-variable constructor path.

soulomoon (commented):
LGTM

Reviewed commit: `f6d8314027414ad5c65bd11bb246ce84dd4aec7b`

PR #35, opened 2026-04-24
Implement #13

### LINKED ISSUE #13: Investigate whether MLFP modules can be compiled separately
## Context

We need to determine whether modules in MLFP can be compiled separately, rather than requiring whole-program compilation.

## Questions to answer

- What does the thesis-backed design expect for MLFP module compilation?
- Do current module boundaries preserve enough information for separate compilation?
- Which phases currently assume whole-program visibility?
- If separate compilation is intended, what interface artifact or summary should each compiled module produce?
- If separate compilation is not intended, where should that constraint be documented?

## Expected outcome

Document the intended behavior and, if needed, add tests or implementation work that makes the module compilation model explicit and thesis-faithful.

soulomoon:
Closes #13.

## Implementation Plan

---
issue_number: 13
pr_number: 35
branch: codex/issue-13
---

## Decision

Implement issue #13 as an explicit documentation-and-regression slice: `.mlfp` modules are same-compilation-unit modules today. Separate compilation is not supported until a future compiler-owned interface artifact is designed around exported principal schemes, visible type/class/instance summaries, and checked xMLF/runtime payloads.

## Plan

1. Before editing, synchronize `codex/issue-13` with `origin/HEAD`. The branch currently has an empty starter commit and is behind current `origin/master`; rebase onto `origin/HEAD` or otherwise update the branch so the PR does not accidentally revert newer base changes. Verify `git status --short` is clean and PR #35 still uses head `codex/issue-13`.
2. Update `docs/mlfp-language-reference.md` under `Modules` to state the supported contract clearly: a `.mlfp` `Program` is one compilation unit; imports resolve only against modules in that same parsed program; modules may be declared later in that program; the CLI runner adds the built-in Prelude explicitly; there is no filesystem module discovery, persisted interface file, stable `.mlfp` ABI, linker, or separate compilation mode yet.
3. Mirror the short user-facing version in `README.md` near the unified `.mlfp` surface summary, pointing readers to the language reference for details.
4. Update `docs/architecture.md` in the `.mlfp` frontend/program ownership area to record the implementation boundary: `Resolve`/`Check` topologically sort the full `Program`, `Finalize` consumes the assembled program scope, and `Run` evaluates all checked module bindings together. Note that future separate compilation would need an explicit interface artifact rather than source peeking.
5. Add a concise `CHANGELOG.md` Unreleased entry because this settles a user-visible module-compilation contract.
6. Add focused regression coverage in `test/ProgramSpec.hs`, preferably under `MLF.Program diagnostics`, with a test named around same-compilation-unit imports. Parse a program that imports a module absent from that program and assert `checkProgram program == Left (ProgramUnknownImportModule "ExternalCore")`. Keep it in-memory unless a CLI filesystem search regression appears necessary; current behavior is already enforced by `topoSortModules`/`buildImportScope` and `runProgramFile` parsing one file plus explicit Prelude.
7. Avoid production code changes unless the new test exposes behavior contrary to the intended contract. If production code is touched, keep it narrowly within `MLF.Frontend.Program.*` and preserve the existing fail-closed unknown-import diagnostic.
8. Validate with a focused ProgramSpec run, for example `cabal test mlf2-test --test-show-details=direct --test-options='--match "MLF.Program diagnostics"'`, then run `cabal build all && cabal test` before handoff because the final diff includes regression coverage.
9. Before publishing, inspect `git status --short` and relevant diffs, stage only issue-related docs/tests, commit with an imperative message as `codex-watcher <codex-watcher@users.noreply.github.com>`, run `gh auth status` and `gh auth setup-git`, then push the existing `codex/issue-13` branch for PR #35. Do not create a new PR or resolve PR review threads.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## Unreleased
 
 ### Changed
+- Documented the `.mlfp` module compilation contract: modules are checked as
+  one parsed `Program` compilation unit, the CLI Prelude is added explicitly,
+  and separate compilation/interface files remain future work. Added regression
+  coverage for imports that name modules outside the current program.
 - Enforced the `.mlfp` resolved-symbol phase boundary. Program syntax is now
   phase-indexed, the resolver produces a resolved AST with semantic symbols at
   global reference sites and resolved type heads, `checkProgram` routes through
--- a/README.md
+++ b/README.md
@@ -91,6 +91,12 @@ The unified `.mlfp` surface includes:
 - single-parameter typeclasses, class constraints, and schema instances
 - `deriving Eq` for nullary, recursive, and parameterized ADTs whose fields have Eq evidence
 
+`.mlfp` modules are same-compilation-unit modules today: imports resolve among
+modules in the parsed program, with the CLI runner adding the built-in Prelude
+as an explicit module. Separate module compilation and interface files are not
+supported yet; see `docs/mlfp-language-reference.md` for the user-facing module
+contract.
+
 Internally, `.mlfp` now reuses the old MLF ownership boundary:
 
 - `MLF.Frontend.Program.Check` assembles module/import/class/data environments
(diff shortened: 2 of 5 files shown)


soulomoon (commented):
LGTM

Reviewed commit: `b1e2ddd6f404bbfd33a89465ccd8c58ff770d98c`

Issue #18, opened 2026-04-22
Elaborate and test higher-kinded .mlfp classes and data declarations


soulomoon:
Parent: #8

Scope:
Elaborate checked higher-kinded .mlfp classes and data declarations into the existing downstream representation and add end-to-end coverage.

Acceptance criteria:
- Well-kinded higher-kinded class and data declarations elaborate successfully.
- Invalid programs rejected by #17 are not accepted through elaboration shortcuts.
- End-to-end tests exercise representative class, data, and instance scenarios.
- Existing elaboration behavior for first-order programs remains unchanged.

Dependencies / compatibility:
Depends on #15, #16, and #17. This sub-issue should consume the checked representation from siblings instead of redefining parser or kind-checker logic.

Issue #17, opened 2026-04-22
Kind-check higher-kinded .mlfp declarations and instance heads


soulomoon:
Parent: #8

Scope:
Implement kind-checking rules for higher-kinded .mlfp declarations and instance heads, using the syntax/model work from #15 and #16.

Acceptance criteria:
- Kind checker validates higher-kinded declaration parameters and type constructor applications.
- Instance heads reject kind mismatches and arity mismatches with actionable diagnostics.
- Existing first-order declarations and instances continue to pass unchanged.
- Tests cover valid higher-kinded use, invalid arity, invalid kind, and mixed first-order/higher-kinded cases.

Dependencies / compatibility:
Depends on #15 and #16. Avoid elaboration/runtime changes reserved for #18; this issue should only decide whether source declarations are well-kinded.

Issue #16, opened 2026-04-22
Parse and pretty-print variable-headed .mlfp type application


soulomoon:
Parent: #8

Scope:
Add parser and pretty-printer support for variable-headed .mlfp type application, using the higher-kinded source model from #15.

Acceptance criteria:
- Parser accepts valid variable-headed type applications involving higher-kinded parameters.
- Parser rejects ambiguous or malformed type applications with useful diagnostics.
- Pretty-printer preserves the intended type application structure.
- Round-trip tests cover nested applications and existing first-order syntax.

Dependencies / compatibility:
Depends on #15. Do not duplicate kind-checking rules from #17; this sub-issue should only establish syntax-level structure and round-trip behavior.

PR #36, opened 2026-04-24
Implement #15

### LINKED ISSUE #15: Define the .mlfp higher-kinded source type and kind model
Parent: #8

Scope:
Define the source-level representation for higher-kinded type parameters in .mlfp, including the AST shape, kind representation, and naming rules needed by parser and checker phases.

Acceptance criteria:
- The source model can represent type constructors that accept type parameters.
- The kind model distinguishes first-order types from higher-kinded type constructors.
- Existing first-order .mlfp programs remain representable without migration.
- Unit or golden tests cover representative higher-kinded declarations and existing first-order declarations.

Dependencies / compatibility:
This is the foundation for #16, #17, and #18. Keep the model boundary narrow so parser, kind-checker, and elaboration work can be implemented independently by sibling sub-issues.

soulomoon:
Closes #15.

## Implementation Plan

---
issue_number: 15
pr_number: 36
branch: codex/issue-15
---

## Implementation Plan

1. Re-verify implementation context before editing: stay on `codex/issue-15`, confirm PR #36 still points at `codex/issue-15` into `master`, and check `git status --short` for unrelated dirty files.

2. Add the source model in `src/MLF/Frontend/Syntax.hs` and re-export it through `MLF.Frontend.Syntax.Program` / public shims as needed:
   - Define `SrcKind = KType | KArrow SrcKind SrcKind` (or equivalent names) for `*`, `* -> *`, and higher arities by right-association.
   - Define `TypeParam` with `typeParamName :: String` and `typeParamKind :: SrcKind`, plus helpers such as `firstOrderTypeParam`, `typeParamNames`, and kind/default accessors.
   - Extend `SrcTy` and `ResolvedSrcTy` with an explicit variable-headed type application constructor, e.g. `STVarApp String (NonEmpty SrcType)` / `RSTVarApp String (NonEmpty ResolvedSrcType)`, so `f a` can be represented without treating `f` as an uppercase/global type constructor. Update `resolvedSrcTypeToSrcType`, `resolvedSrcTypeIdentityType`, recursion-scheme base functor support, pattern synonyms/complete pragmas if needed, normalization, free-variable, and substitution helpers to recurse through the new constructor.

3. Thread typed parameters through the `.mlfp` declaration AST in `src/MLF/Frontend/Syntax/Program.hs`:
   - Change `classDeclParam` and `dataDeclParams` to carry `TypeParam` values.
   - Use helper functions where downstream code only needs names, keeping existing first-order behavior equivalent to all params defaulting to `KType`.
   - Update `unresolve*` functions and resolved syntax types without changing module/import/export semantics.

4. Add parser support for declaration parameter kind annotations, but do not implement #16 variable-headed type-application parsing yet:
   - In `src/MLF/Frontend/Parse/Program.hs`, parse bare lowercase params as `firstOrderTypeParam name`.
   - Parse parenthesized kind annotations such as `(f :: * -> *)`, with right-associative kind arrows and bare `*` for first-order types.
   - Apply the same grammar to `class` parameters and `data` parameters.
   - Keep existing first-order `.mlfp` syntax unchanged.

5. Update pretty-printing and docs:
   - In `src/MLF/Frontend/Pretty/Program.hs`, pretty-print first-order params as bare names and higher-kinded params as `(name :: kind)`.
   - In `src/MLF/Frontend/Pretty.hs`, add rendering for variable-headed type application so directly constructed AST values have a stable representation for #16.
   - Update `docs/mlfp-language-reference.md` to document declaration parameter kind annotations/default `*`, and explicitly note that full variable-headed type-application parsing/kind checking/elaboration are handled by follow-up issues #16-#18.

6. Add narrow checker/model validation without implementing full kind checking from #17:
   - Ensure duplicate data type parameter names are rejected with a clear `ProgramError` and diagnostic span where feasible.
   - Preserve current constructor-result arity validation by comparing against `length (typeParamNames dataDeclParams)`.
   - Update `DataInfo`, `ClassInfo`, deriving, instance skeletons, constructor metadata, and elaboration helpers to use parameter names through helpers while retaining parameter kind metadata for later phases.
   - For any variable-headed type applications that reach downstream phases before #17/#18, keep behavior fail-closed and explicit rather than silently lowering them to named `TCon`s.

7. Add tests focused on #15 boundaries:
   - Parser/pretty round-trip for existing first-order declarations to prove no source migration is needed.
   - Parser/pretty round-trip for `class Functor (f :: * -> *)` and `data Higher (f :: * -> *) a = ...` using signatures that do not require #16 parsing.
   - Direct AST/unit tests showing representative higher-kinded method/constructor types can be represented with the new variable-headed application constructor, e.g. `f a` and `p a b`.
   - Negative test for duplicate data type parameter names if the new checker validation is added.

8. Validation and publish for the implementation turn:
   - Run the narrow relevant slice first, e.g. `cabal test mlf2-test --test-options '--match "MLF.Program"'` plus parser/pretty specs if touched.
   - Run the full required gate before completion: `cabal build all && cabal test`.
   - Inspect `git status --short` and relevant diffs, stage only issue-related source/docs/tests, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, push `codex/issue-15`, and leave PR #36 ready for review without resolving review threads.

---

### DIFF
--- a/src/MLF/Elab/Run/Pipeline.hs
+++ b/src/MLF/Elab/Run/Pipeline.hs
@@ -142,6 +142,8 @@ validateDirectRecursiveAnnotations = goExpr
             Surface.STArrow dom cod -> bodyType True shadowed dom && bodyType True shadowed cod
             Surface.STBase _ -> True
             Surface.STCon _ args -> all (bodyType True shadowed) args
+            Surface.STVarApp v args ->
+              (shadowed || v /= needle || guarded) && all (bodyType True shadowed) args
             Surface.STForall v mb body ->
               let shadowed' = shadowed || v == needle
                   boundOk = maybe True (bodyBound guarded shadowed' . Surface.unNormBound) mb
@@ -156,6 +158,8 @@ validateDirectRecursiveAnnotations = goExpr
             Surface.STArrow dom cod -> bodyType True shadowed dom && bodyType True shadowed cod
             Surface.STBase _ -> True
             Surface.STCon _ args -> all (bodyType True shadowed) args
+            Surface.STVarApp v args ->
+              (shadowed || v /= needle || guarded) && all (bodyType True shadowed) args
             Surface.STForall v mb body ->
               let shadowed' = shadowed || v == needle
                   boundOk = maybe True (bodyBound guarded shadowed' . Surface.unNormBound) mb
@@ -639,6 +643,8 @@ srcTypeToElabType ty = case ty of
   Surface.STArrow dom cod -> TArrow (srcTypeToElabType dom) (srcTypeToElabType cod)
   Surface.STBase name -> TBase (BaseTy name)
   Surface.STCon name args -> TCon (BaseTy name) (fmap srcTypeToElabType args)
+  Surface.STVarApp name _ ->
+    error ("variable-headed source type application `" ++ name ++ "` cannot be lowered before higher-kinded elaboration is implemented")
   Surface.STForall name mb body ->
     TForall name (mb >>= srcBoundToElabBound) (srcTypeToElabType body)
   Surface.STMu name body -> TMu name (srcTypeToElabType body)
@@ -652,6 +658,8 @@ srcTypeToElabType ty = case ty of
       Surface.STArrow dom cod -> Just (TArrow (srcTypeToElabType dom) (srcTypeToElabType cod))
       Surface.STBase name -> Just (TBase (BaseTy name))
       Surface.STCon name args -> Just (TCon (BaseTy name) (fmap srcTypeToElabType args))
+      Surface.STVarApp name _ ->
+        error ("variable-headed source type application `" ++ name ++ "` cannot be lowered before higher-kinded elaboration is implemented")
       Surface.STForall name mb body ->
         Just (TForall name (mb >>= srcBoundToElabBound) (srcTypeToElabType body))
       Surface.STMu name body -> Just (TMu name (srcTypeToElabType body))
--- a/src/MLF/Elab/Run/Pipeline.hs
+++ b/src/MLF/Elab/Run/Pipeline.hs
@@ -228,6 +228,7 @@ runPipelineElabWith ::
   Either PipelineError PipelineElabDetailedResult
 runPipelineElabWith requireFinalTypeCheck traceCfg extBindings genConstraints expr = do
   () <- fromConstraintError (validateDirectRecursiveAnnotations expr)
+  initialSchemeInfos <- fromConstraintError (traverse externalBindingSchemeInfo extBindings)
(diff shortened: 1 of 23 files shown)


src/MLF/Elab/Run/Pipeline.hs
```diff
@@ -639,6 +643,8 @@ srcTypeToElabType ty = case ty of
   Surface.STArrow dom cod -> TArrow (srcTypeToElabType dom) (srcTypeToElabType cod)
   Surface.STBase name -> TBase (BaseTy name)
   Surface.STCon name args -> TCon (BaseTy name) (fmap srcTypeToElabType args)
+  Surface.STVarApp name _ ->
+    error ("variable-headed source type application `" ++ name ++ "` cannot be lowered before higher-kinded elaboration is implemented")
```
soulomoon: This makes unsupported `STVarApp` external binding types crash the public pipeline instead of returning a typed `PipelineError`. The converter is used at line 262 for `runPipelineElabWithEnv` / `runPipelineElabDetailedWithExternalBindings`, before the constraint generator's `InternalConstraintError` path can reject unsupported variable-headed applications. Since `NormSrcType` now publicly admits `STVarApp` and `normalizeType` preserves it, a direct-AST caller can pass an external type like `f a` and hit this `error`. Please prevalidate external bindings or make this conversion return through `ConstraintError`/`PipelineError` so unsupported HK types fail closed.

soulomoon (commented):
Found one additional issue while rechecking the prior threads.

src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -351,14 +361,18 @@ sourceForallMatches expected actual =
   where
     match bound subst template actualTy =
       case template of
-        STForall name _ body -> match (Set.insert name bound) subst body actualTy
+        STForall name _ body ->
+          case actualTy of
+            STForall actualName _ actualBody ->
```
soulomoon: This alpha-renamed forall branch drops both bounds, so `sourceForallMatches` accepts incompatible bounded foralls as long as the bodies match after renaming. For example, an expected `forall (f >= Int). f -> f` and an actual `forall (g >= Bool). g -> g` would reach this branch, ignore the `Int`/`Bool` bounds, and return `True`. Since `finalizeBinding` uses this as a final compatibility fallback after the elaboration-level comparisons fail, incompatible annotations can be accepted instead of producing `ProgramTypeMismatch`; please compare the bounds under the same renamed binder mapping before matching the bodies.

soulomoon (commented):
Found one bounds-handling issue in the new forall matching fallback.

soulomoon (commented):
No actionable findings. The prior bounds-handling review thread is fixed; tests were not run locally because `cabal` is unavailable.

Reviewed commit: `ae53913a62489b7adb3c516cf122a9d819bf1878`

PR #40, opened 2026-04-25
Implement #18

### LINKED ISSUE #18: Elaborate and test higher-kinded .mlfp classes and data declarations

soulomoon:
Closes #18.

## Implementation Plan

---
issue_number: 18
pr_number: 40
branch: codex/issue-18
---

## Implementation Plan

1. **Reconcile branch/worktree before editing.** PR #40 is open from `codex/issue-18` to `master`, but the local branch is currently 1 commit ahead and 46 commits behind `origin/HEAD`, with dirty edits in frontend/program files. Treat the dirty diff as stale reference only unless re-inspection proves otherwise: relative to `origin/HEAD` it removes the #15-#17 `STVarApp`/`TypeParam` higher-kinded model. Preserve the `origin/HEAD` representation and kind checker, and block if dirty-change ownership is unclear.

2. **Implement on top of the checked `STVarApp` model.** Keep parser/pretty/kind-check behavior from #15-#17. Add or centralize a small STVarApp-aware source-type helper for applying a substituted type head to arguments, preserving these cases: `STVar f` + args -> `STVarApp f args`, `STBase T` + args -> `STCon T args`, existing `STCon`/`STVarApp` heads append their existing arguments.

3. **Make substitution and lowering consistently higher-kinded-aware.** Update the existing source-type substitution paths in `Program.Types`, `Program.Elaborate`, and any local finalization/check helpers so substitutions through `STVarApp f args` replace `f` with the concrete or partially applied head instead of leaving unsupported variable-headed applications in specialized instance/method/constructor types.

4. **Extend matching where elaboration resolves concrete uses.** In `Program.Elaborate` and `Program.Check`, teach instance/type matching and overlap checks to match template `STVarApp f args` against concrete `STCon Head (prefix ++ args)` or variable-headed applications, binding `f` to the recovered head while retaining occurs checks and alpha/semantic equality. Keep unresolved `STVarApp` fail-fast at the final `srcTypeToElabType` boundary so invalid or still-polymorphic higher-kinded uses do not bypass #17.

5. **Keep recovery/finalization aligned.** Ensure `Program.Finalize` source-forall matching, constructor/data recovery, and `matchRecoverType` continue to recover partially applied constructor heads for higher-kinded data declarations without converting unresolved higher-kinded variables into first-order downstream types.

6. **Add end-to-end ProgramSpec coverage.** Add representative `.mlfp` tests for: a higher-kinded class method instantiated with a parameterized data constructor, a higher-kinded data declaration with a variable-headed field that checks/elaborates when used at a concrete head, and at least one invalid #17-style arity/kind program exercised through the normal program check/run path to prove elaboration does not shortcut kind validation.

7. **Validate in slices, then full gate.** First run focused tests around `higher-kinded`, `MLF.Program source type finalization`, and the relevant program matrix cases. Then run the required behavior-changing gate: `cabal build all && cabal test`.

8. **Publish safely.** Before committing, run `git status --short` and inspect relevant diffs. Stage only issue-related source/test files, never watcher state. Run `gh auth status` and `gh auth setup-git`, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, push `codex/issue-18`, and leave PR #40 ready for watcher/review handoff without creating another PR.

---


src/MLF/Frontend/Program/Check.hs
```diff
@@ -1660,9 +1661,27 @@ addConstructorValues moduleName0 dataInfos =
...
 constructorRuntimeBindingRecoverable :: ConstructorInfo -> Bool
 constructorRuntimeBindingRecoverable ctor =
-  let evidenceVars = foldMap freeTypeVars (ctorArgs ctor ++ [ctorResult ctor])
-   in all (\(name, _) -> name `Set.member` evidenceVars) (ctorForalls ctor)
+  let involvedTypes =
+        ctorArgs ctor
+          ++ [ctorResult ctor]
+          ++ mapMaybe snd (ctorForalls ctor)
+      evidenceVars = foldMap freeTypeVars involvedTypes
+   in not (any hasVariableHeadApplication involvedTypes)
```
soulomoon: This check only looks at the constructor being bound, but lowering a constructor binding builds the full Church handler chain for every constructor of the owning data type. A mixed higher-kinded ADT such as `data MaybeF (f :: * -> *) a = NothingF : MaybeF f a | JustF : f a -> MaybeF f a;` still marks `NothingF` recoverable/trackable, so its generated binding/runtime type includes the `JustF` handler type `f a` and the pipeline fails with `variable-headed source type application 'f' is not supported` before the program can run. The recoverability/trackability decision needs to be owner-wide and kept consistent with the copies in `Elaborate` and `Finalize`, or nullary constructors from such data types need to be inlined too.

soulomoon (commented):
Found one issue with mixed-constructor higher-kinded ADTs.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -976,6 +1015,33 @@ compileResolvedValueApp scope mbExpected valueInfo args = do
         Nothing -> pure ()
       compileResolvedExpr scope (Just expectedTy) arg
 
+compileConstructorHead :: ElaborateScope -> ConstructorInfo -> Int -> Map String SrcType -> ElaborateM SurfaceExpr
+compileConstructorHead scope ctorInfo argCount constructorSubst = do
+  placeholder <- deferConstructorCall scope ctorInfo argCount constructorSubst
```
soulomoon, line 1045: This still defers owner-untrackable higher-kinded constructors through an external placeholder whose type is lowered immediately in `deferConstructorCall`. If the constructor is first-class before the concrete result type reaches the constructor itself, the substitution is empty and the placeholder type still contains `f a`, so the pipeline fails before `inlineConstructorHead` can recover it. For example, with `data MaybeF (f :: * -> *) a = NothingF : MaybeF f a | JustF : f a -> MaybeF f a`, `def id : forall a. a -> a = \x x; def main : MaybeF Box Bool = id NothingF;` currently errors with `variable-headed source type application 'f' is not supported`. Owner-untrackable constructors need to avoid registering an unresolved lowered placeholder type, or be inlined before the pipeline sees the placeholder.

soulomoon (commented):
Found one remaining higher-kinded constructor regression.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -1057,6 +1073,27 @@ applyResolvedImportItem moduleName0 exports scope item =
           pure (mkScope values types (scopeClasses scope) (scopeInstances scope))
         Nothing -> throwError (ProgramImportNotExported moduleName0 (P.resolvedExportTypeName ref))
 
+importConstructorOwnerType :: P.ModuleName -> ModuleExports -> Scope -> ValueInfo -> TcM Scope
+importConstructorOwnerType _moduleName0 exports scope = \case
+  ConstructorValue {valueCtorInfo = ctorInfo} ->
+    case exportedConstructorOwnerType ctorInfo exports of
+      Just dataInfo -> do
+        let hiddenName = ctorOwningType ctorInfo
+        types <- liftEither (addTypes (scopeTypes scope) (Map.singleton hiddenName dataInfo))
```
soulomoon: This hidden owner import is still inserted into the visible type-name map under `ctorOwningType`. For a value-only constructor import, the resolver has not exposed that type name, so `module Main` can legally declare its own `data MaybeF ...`; after `import Core exposing (NothingF)`, this internal insert makes the later `addTypes (scopeTypes importScope) localData` fail with a duplicate visible name even though `MaybeF` was never imported. Please keep the owner metadata out of the visible `scopeTypes` namespace (for example by carrying it by identity/private key) so constructor-value imports do not reserve or expose the owner's source name.

soulomoon (commented):
Found one import-scope regression.

soulomoon (commented):
Found one remaining constructor export edge case.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -1054,9 +1090,105 @@ applyResolvedImportItem moduleName0 exports scope item =
...
           types <- liftEither (addTypes (scopeTypes scope) (Map.singleton typeName dataInfo))
-          pure (mkScope values types (scopeClasses scope) (scopeInstances scope))
+          pure (mkScopeWithHidden values types (scopeHiddenTypes scope) (scopeClasses scope) (scopeInstances scope))
         Nothing -> throwError (ProgramImportNotExported moduleName0 (P.resolvedExportTypeName ref))
 
+prepareImportedValue :: ModuleExports -> Scope -> ValueInfo -> (Scope, ValueInfo)
+prepareImportedValue exports scope valueInfo =
+  case valueInfo of
+    ConstructorValue {valueCtorInfo = ctorInfo} ->
+      case exportedConstructorOwnerType ctorInfo exports of
```
soulomoon, line 1118: This still only creates a hidden owner when the owning type is also present in `exportedTypes`. A module can explicitly export a constructor value without exporting its type, e.g. `module Core export (Box(..), NothingF, accept) { data MaybeF (f :: * -> *) a = NothingF : MaybeF f a | JustF : f a -> MaybeF f a; ... }`; importing `NothingF` then leaves no `DataInfo` for `MaybeF` in the consumer. The later inline path has `ctorOwnerConstructors`, but `inlineConstructorHead` lowers `ctorResult` without the owner in `dataTypes`, so it builds `ERoll` over an ordinary `TCon "MaybeF"` and final type checking fails with `TCExpectedRecursive`. Please carry/register hidden owner metadata for constructor-value exports even when the type itself is not exported.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -1054,9 +1091,218 @@ applyResolvedImportItem moduleName0 exports scope item =
...
+
+constructorOwnerParamNames :: ConstructorInfo -> [String]
+constructorOwnerParamNames ctorInfo =
+  case ctorResult ctorInfo of
+    STBase name
+      | name == ctorOwningType ctorInfo || name == symbolDefiningName (ctorOwningTypeIdentity ctorInfo) ->
+          []
+    STCon name args
+      | name == ctorOwningType ctorInfo || name == symbolDefiningName (ctorOwningTypeIdentity ctorInfo) ->
+          mapMaybe srcTypeVarName (NE.toList args)
```
soulomoon: This reconstructs the owner's parameter list from only the imported constructor result and drops any result argument that is not a bare type variable. GADT-style results are accepted elsewhere, so a value-only export such as `data MaybeF (f :: * -> *) a = NothingBool : MaybeF f Bool | JustF : f a -> MaybeF f a` would rebuild the hidden owner with only `f` in `dataParams`; `lowerType` then zips only that parameter and leaves the sibling `JustF` handler's `a` free. Please carry the declaring `dataTypeParams`/`dataParams` (or enough owner-shape metadata to reconstruct all declared parameters) instead of inferring them from one constructor result.

soulomoon (commented):
Found one actionable issue in the new hidden-owner reconstruction path.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -585,7 +609,7 @@ addAllExports scope exports =
     values <- addValues (scopeValues scope) (exportedValues exports)
     types <- addTypes (scopeTypes scope) (Map.map exportedTypeData (exportedTypes exports))
     classes <- addClasses (scopeClasses scope) (exportedClasses exports)
-    pure (mkScope values types classes (scopeInstances scope))
+    pure (mkScopeWithHidden values types (scopeHiddenTypes scope) classes (scopeInstances scope))
```
soulomoon: Carrying `scopeHiddenTypes` through here is not enough because `addAllExports` still inserts exported constructor values unchanged. For a module that exports `(Box(..), NothingF, accept)` but not `MaybeF`, a consumer using `import Core` or `import Core as C` gets `NothingF`/`C.NothingF` without `prepareImportedValue`, so no hidden `$Core.MaybeF` owner is registered and the inline constructor path still lowers the result as an ordinary `TCon "MaybeF"`. Please run constructor values imported through this bulk path through the same hidden-owner preparation used by explicit exposing imports.

soulomoon (commented):
Found one remaining import-path hole in the hidden owner metadata handling.

src/MLF/Frontend/Program/Check.hs
```diff
@@ -580,12 +604,29 @@ resolvedImportDefiningModule =
...
+    pure (mkScopeWithHidden values types (scopeHiddenTypes scopeWithOwners) classes (scopeInstances scopeWithOwners))
+
+prepareBulkImportedValues :: ModuleExports -> Scope -> Map String ValueInfo -> (Scope, Map String ValueInfo)
+prepareBulkImportedValues exports =
+  Map.mapAccumWithKey (\scope0 _ valueInfo -> prepareBulkImportedValue exports scope0 valueInfo)
+
+prepareBulkImportedValue :: ModuleExports -> Scope -> ValueInfo -> (Scope, ValueInfo)
+prepareBulkImportedValue exports scope valueInfo@ConstructorValue {valueCtorInfo = ctorInfo}
+  | constructorOwnerVisibleInExports ctorInfo exports = (scope, valueInfo)
+  | otherwise = prepareImportedValue exports scope valueInfo
```
soulomoon, line 622: This prepares each value-only constructor independently and overwrites the same hidden owner entry as the map traversal proceeds. For an aliased bulk import that exports two constructors but not the owner type, e.g. `module Core export (Box(..), NothingF, JustF) ...` then `import Core as C`, the final hidden `DataInfo` is built from only the last processed constructor. Its sibling constructor symbols are synthesized from qualified shape names like `C.JustF`, so the earlier imported value still carries the real `JustF` symbol and no longer belongs to the hidden owner in `resolveConstructorDataInfo`; a `case C.JustF ... of { C.NothingF -> ...; C.JustF box -> ... }` can fail with an unknown owner. Please group/reuse hidden owner metadata per owner identity, or carry the original sibling constructor identities, so all imported constructors from the same hidden owner point at the same `DataInfo`.

soulomoon (commented):
Found one remaining edge case in the bulk constructor-owner preparation.

soulomoon (commented):
Clean review at 0bddbb54d1594cedf8602d0ad8b66376d09ae0ce.

Reviewed commit: `0bddbb54d1594cedf8602d0ad8b66376d09ae0ce`

PR #42, opened 2026-04-25
Implement #8

### LINKED ISSUE #8: Support higher-kinded type parameters in .mlfp
This is valid as an enhancement / design task for `.mlfp`, not as a current-contract bug.

## Goal

`.mlfp` should be able to express higher-kinded/type-operator abstractions so common interfaces can be encoded directly, for example:

```mlfp
class Functor f {
  fmap : forall a b. (a -> b) -> f a -> f b;
}

class Monad m {
  pure : forall a. a -> m a;
  bind : forall a b. m a -> (a -> m b) -> m b;
}

class Profunctor p {
  dimap : forall a b c d. (a -> b) -> (c -> d) -> p b c -> p a d;
}
```

A higher-kinded data declaration should also be representable, e.g.:

```mlfp
data Higher f a =
    MkHigher : f a -> Higher f a;
```

## Current Behavior

The current `.mlfp` frontend rejects these forms before type checking. Type application only works when the head is an uppercase named type constructor. Lowercase type variables parse as atomic `STVar`s and cannot be applied to arguments, so `f a`, `m a`, and `p a b` are not representable today.

Relevant code paths:

- `src/MLF/Parse/Type.hs`: `parseArrowTypeWith` builds `STCon` only when the type application head parses as an uppercase constructor.
- `src/MLF/Frontend/Parse/Program.hs`: data parameters are currently parsed as plain lowercase names with no kind information.
- `src/MLF/Frontend/Program/Check.hs`: constructor-result validation checks owning data head/arity, but there is no kind checking for type-operator parameters.

I verified this with local CLI probes on 2026-04-21:

```mlfp
class Functor f {
  fmap : forall a b. (a -> b) -> f a -> f b;
}
```

and

```mlfp
data Higher f a =
    MkHigher : f a -> Higher f a;
```

Both are rejected by parsing. A first-order class such as `class Idd a { keep : a -> a; }` works, so the missing capability is specifically higher-kinded/type-operator application, not typeclasses in general.

## Thesis Context

This does not appear to be required by the current thesis-faithfulness contract. The thesis source treats higher-order types as future work and notes that extending MLF to higher-order types is non-trivial. So this should be tracked as a scoped `.mlfp` extension rather than a defect in the current MLF implementation.

## Design Work Needed

Likely pieces:

- Add an explicit kind story for source type parameters, at least enough for `*`, `* -> *`, and `* -> * -> *`.
- Extend source type syntax/AST so type application can have a variable head, e.g. `f a`, instead of only uppercase `STCon` heads.
- Decide whether the implementation should use a limited higher-order-kinds encoding with no type-level reduction, or require explicit higher-kind abstractions/applications.
- Add kind checking for data declarations, class declarations, instance heads, constructor signatures, and constrained method types.
- Update elaboration/lowering and instance resolution so higher-kinded parameters remain coherent and fail closed.
- Add positive and negative `.mlfp` corpus tests for `Functor`, `Monad`, `Profunctor`, and a higher-kinded data declaration.
- Update `docs/mlfp-language-reference.md` once the supported subset is decided.

## Related Subtasks

- [ ] #9 tracks source-level `Monoid` support through expected-type resolution for nullary overloaded methods such as `mempty : a`.

soulomoon:
Closes #8.

## Implementation Plan

---
issue_number: 8
pr_number: 42
branch: codex/issue-8
---

## Implementation Plan

1. Start from the prepared branch `codex/issue-8` and re-check `git status --short --branch`, `gh issue view 8`, and `gh pr view 42` to confirm the branch, issue, and existing PR are still the watcher-owned targets.
2. Treat the current codebase as already containing the intended issue #8 implementation unless verification disproves it. Evidence inspected during planning: `SrcKind`, `TypeParam`, and `STVarApp` exist; `.mlfp` parsing accepts kinded params and variable-headed type apps; program checking validates kinds for data/class declarations, constraints, methods, definitions, and instance heads; docs and changelog already describe the supported higher-kinded subset.
3. Run focused validation first: the higher-kinded and variable-headed `.mlfp` examples in `test/ProgramSpec.hs`, plus related parser/pretty/normalize coverage in `FrontendParseSpec`, `FrontendPrettySpec`, and `FrontendNormalizeSpec`. Planning attempted a narrow `cabal test` slice, but `cabal` was not on `PATH`; if that remains true in implementation, return blocked rather than claiming validation.
4. If focused validation passes and `git diff origin/HEAD...HEAD` remains empty except the watcher starter commit, make no code changes. The implementation is already present on the base branch; the later turn should report completion after full required validation/publish checks.
5. If validation exposes a real gap, patch only the relevant issue #8 surface: `src/MLF/Parse/Type.hs`, `src/MLF/Frontend/Syntax.hs`, `src/MLF/Frontend/Parse/Program.hs`, `src/MLF/Frontend/Pretty/Program.hs`, `src/MLF/Frontend/Program/Resolve.hs`, `src/MLF/Frontend/Program/Check.hs`, and related elaboration/finalization paths that preserve `STVarApp` until checked/lowered.
6. Keep the supported design scoped: explicit `*` and right-associative `* -> ...` source kinds, kinded data/class parameters, variable-headed type applications such as `f a` and `p a b`, fail-closed handling for raw pipeline `STVarApp`, and no kind polymorphism, type lambdas, or type-level reduction.
7. Maintain/update regression coverage for positive `Functor`, `Monad`, `Profunctor`, and higher-kinded data declarations, plus negative cases for first-order parameters used as type functions, unsaturated/over-applied constructors, and mismatched higher-kinded arguments.
8. Update `docs/mlfp-language-reference.md` and `CHANGELOG.md` only if validation-driven code changes alter the documented supported subset or user-visible behavior.
9. Before publishing from an implementation turn, run `gh auth status` and `gh auth setup-git`, inspect `git status --short` and relevant diffs, stage only issue-related files, commit as the prepared `codex-watcher` identity if there are changes, push `codex/issue-8`, and avoid creating a duplicate PR.
10. Final validation target before handoff is `cabal build all && cabal test` for behavior-changing changes. If no code changes are needed because issue #8 is already implemented on base, still run the strongest available validation and explicitly report any environment blocker such as missing `cabal`.

---


soulomoon (commented):
Clean review: PR #42 has no changed files against master at head 3878f3247f2ecf012de1a45ac135f7d9fff21d1d, so there are no actionable inline findings.

Reviewed commit: `3878f3247f2ecf012de1a45ac135f7d9fff21d1d`

PR #45, opened 2026-04-26
Implement #19

### LINKED ISSUE #19: Define typed backend IR invariants and module boundary
Parent: #10

Scope:
Define the typed backend IR module boundary and the invariants that later conversion and lowering steps must preserve.

Acceptance criteria:
- A backend IR module/type structure exists with explicit types for values, functions, and module-level declarations.
- Core invariants are documented in code or tests, including type preservation and valid reference boundaries.
- The IR can represent a minimal checked .mlfp program without committing to LLVM lowering details.
- Tests validate construction or decoding of representative valid/invalid IR shapes.

Dependencies / compatibility:
This is the foundation for #20 and #21. Keep LLVM textual concerns out of this issue so sibling work can depend on a stable typed IR contract.

soulomoon:
Closes #19.

## Implementation Plan

---
issue_number: 19
pr_number: 45
branch: codex/issue-19
---

1. Preserve the existing local scaffold on `codex/issue-19`: `src/MLF/Backend/IR.hs`, `test/BackendIRSpec.hs`, cabal/test wiring, and docs/changelog notes. Do not discard the dirty worktree.
2. Finalize `MLF.Backend.IR` as a private typed backend boundary only: keep LLVM/textual lowering out, expose explicit program/module/data/constructor/binding/type/expression/pattern structures, and keep it under the `mlf2-internal` cabal stanza.
3. Strengthen `validateBackendProgram` around the issue acceptance criteria: global module/binding uniqueness, existing `backendProgramMain`, binding/body type agreement, lexical/global variable reference checks, constructor metadata lookup, constructor arity/argument/result checks, and case alternative result/scrutinee/constructor boundary checks. Keep validation local to the IR contract rather than adding a second source typechecker.
4. Add or extend `BackendIRSpec` coverage for representative valid and invalid shapes: minimal checked-like program, duplicate bindings/modules, missing `main`, unknown variable, invalid application/lambda/let/type application, recursive roll/unroll mismatch, valid ADT construction/case, unknown constructor, constructor arity/type mismatch, and case alternative result mismatch.
5. Sync documentation with the final contract in `docs/architecture.md`, `docs/mlfp-language-reference.md`, `implementation_notes.md`, `README.md`, and `CHANGELOG.md`, making clear this is the backend-owned IR boundary after checked `.mlfp` programs and before lowering.
6. Validate with the narrow slice first, then the full required gate: `cabal build mlf2:mlf2-internal mlf2:mlf2-test`, focused `cabal test mlf2-test --test-options "--match MLF.Backend.IR"`, then `cabal build all && cabal test` before completion.
7. Before publishing, run `gh auth status` and `gh auth setup-git`, inspect `git status --short` and relevant diffs, stage only issue-related files, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, rebase or otherwise reconcile with `origin/HEAD` only once the worktree is clean, push `codex/issue-19`, and leave PR #45 ready for review without creating another PR or resolving review threads.

---


src/MLF/Backend/IR.hs
```diff
@@ -0,0 +1,466 @@
...
+  where
+    go ty =
+      case ty of
+        BTVar name
+          | name == needle -> replacement
+          | otherwise -> ty
+        BTArrow dom cod -> BTArrow (go dom) (go cod)
+        BTBase {} -> ty
+        BTCon con args -> BTCon con (fmap go args)
+        BTForall name mbBound body
```
soulomoon, line 332: `substituteBackendType` skips under same-named binders, but it does not freshen binders that occur free in the replacement. That makes `BackendTyApp` capture type variables, e.g. instantiating `forall a. forall b. a -> b` with `BTVar "b"` produces `forall b. b -> b` instead of `forall fresh. b -> fresh`, so the validator can accept or reject the wrong IR. Please make this substitution capture-avoiding, or reject shadowing/unsafe substitutions before using it for type-application validation.

soulomoon (commented):
Reviewed the PR plan and linked issue; I found one correctness issue in the new backend type substitution used by validation.

src/MLF/Backend/IR.hs
```diff
@@ -0,0 +1,521 @@
...
+        Left (BackendLetBodyTypeMismatch resultTy (backendExprType body))
+    BackendTyAbs resultTy name mbBound body -> do
+      validateBackendExprWith mbContext body
+      let expected = BTForall name mbBound (backendExprType body)
+      unless (resultTy == expected) $
+        Left (BackendTypeAbsTypeMismatch name resultTy expected)
+    BackendTyApp resultTy fun tyArg -> do
+      validateBackendExprWith mbContext fun
+      case backendExprType fun of
+        BTForall name _ bodyTy -> do
```
soulomoon: `BackendTyApp` ignores the optional bound on `BTForall`, so an IR node like `forall a >= Int. a -> a` can be instantiated with `Bool` and pass validation as long as the carried result type is `Bool -> Bool`. The frontend checker rejects that shape, and the backend IR keeps the bound in `BackendType`, so the validator should reject type arguments that do not satisfy the explicit bound and add a regression for bounded `BackendTyApp`.

soulomoon (commented):
I found one backend IR validation gap in bounded type applications.

src/MLF/Backend/IR.hs
```diff
@@ -0,0 +1,534 @@
...
+  pure ()
+validateBackendConstructorUse (Just context0) name resultTy args =
+  case Map.lookup name (bvcConstructors context0) of
+    Nothing ->
+      Left (BackendUnknownConstructor name)
+    Just constructor -> do
+      let fields = backendConstructorFields constructor
+      unless (length fields == length args) $
+        Left (BackendConstructorArityMismatch name (length fields) (length args))
+      unless (backendConstructorResult constructor == resultTy) $
```
soulomoon: This exact comparison makes parameterized constructor metadata impossible to validate. For example, `Option a` metadata would naturally give `Some` fields `[BTVar "a"]` and result `BTCon (BaseTy "Option") (BTVar "a" :| [])`, but a checked `Some 1 :: Option Int` node carries result `Option Int`, so it fails here and would also fail the field check (`a` vs `Int`). Since `BackendData` carries parameters and this boundary is meant to support checked `.mlfp` data paths, constructor use/case validation needs to instantiate or match the constructor parameters from the use site, or the IR needs explicit constructor type arguments.

soulomoon (commented):
I found one remaining backend IR validation gap for parameterized ADT constructor metadata.

src/MLF/Backend/IR.hs
```diff
@@ -0,0 +1,731 @@
...
+                    >>= \(substitution', matched) ->
+                      if matched
+                        then Just substitution'
+                        else Nothing
+              where
+                expectedArgsList = NE.toList expectedArgs
+                actualArgsList = NE.toList actualArgs
+            (BTForall expectedName expectedBound expectedBody, BTForall actualName actualBound actualBody) -> do
+              substitution' <- matchMaybeBound bound substitution expectedBound actualBound
+              let used =
```
soulomoon, line 657: The fresh binder set here omits the free variables already recorded in `substitution`. That can make the matcher capture a previously inferred constructor parameter when it compares under a forall. For example, after matching a result that records `p := BTVar "a1"`, a field type like `forall a. p` can be accepted against `forall x. x` because `freshNameLike "a"` may choose `"a1"`, making the free parameter substitution look like the actual bound variable. Please include the free vars of the current substitution values in this `used` set (and the analogous `BTMu` branch) and add a regression so malformed parameterized constructor/case metadata does not pass validation.

soulomoon (commented):
I found one remaining capture issue in the constructor-parameter matcher added for backend ADT validation.

soulomoon (commented):
Clean review: the implementation satisfies the PR plan and linked issue criteria at the reviewed commit, with no actionable findings found.

Reviewed commit: `286d77a6f94f151c05b3031afb85a72ea6e9d269`

Issue #21, opened 2026-04-22
Add first LLVM-like textual lowering boundary for backend IR


soulomoon:
Parent: #10

Scope:
Add the first LLVM-like textual lowering boundary for typed backend IR so backend output can be inspected and tested.

Acceptance criteria:
- A lowering entry point converts supported backend IR modules to deterministic textual output.
- Output preserves function/type structure needed for minimal examples.
- Unsupported IR nodes produce explicit diagnostics or blocked results.
- Golden tests cover at least one minimal module and one unsupported case.

Dependencies / compatibility:
Depends on #19 and should consume conversion output from #20 when available. Do not expand frontend checking here; this issue owns only the IR-to-text boundary.

PR #46, opened 2026-04-26
Implement #20

### LINKED ISSUE #20: Convert checked .mlfp programs to typed backend IR
Parent: #10

Scope:
Convert checked .mlfp programs into the typed backend IR defined by #19.

Acceptance criteria:
- Checked functions and declarations map into backend IR with preserved type information.
- Conversion rejects or reports unsupported checked constructs explicitly instead of silently dropping them.
- Tests cover simple functions, data paths needed by current examples, and unsupported construct reporting.
- The conversion API is reusable by the lowering work in #21.

Dependencies / compatibility:
Depends on #19. Do not implement textual LLVM output here; keep lowering concerns isolated for #21.

soulomoon:
Closes #20.

## Implementation Plan

---
issue_number: 20
pr_number: 46
branch: codex/issue-20
---

## Implementation Plan

1. Synchronize the implementation branch before coding.
   - Confirm `git status --short` is clean.
   - Fetch and rebase `codex/issue-20` onto `origin/HEAD`/`origin/master`; PR #46 currently has only the starter commit and predates merged PR #45 for issue #19.
   - Verify `src/MLF/Backend/IR.hs`, `test/BackendIRSpec.hs`, and the #19 cabal/test wiring are present after rebase.

2. Add an internal backend conversion module.
   - Create `src/MLF/Backend/Convert.hs` and add it to the private `mlf2-internal` exposed modules in `mlf2.cabal`.
   - Keep it internal for now, alongside `MLF.Backend.IR`; do not widen the public `MLF.Pipeline` surface unless implementation shows #21 cannot reuse the private API.
   - Export a small API: `BackendConversionError(..)`, `convertCheckedProgram`, and focused helpers only if tests/lowering need them, such as `convertCheckedBinding` or `convertElabType`.

3. Define fail-closed conversion errors.
   - Include errors for unsupported source/backend type shapes, unsupported xMLF terms/instantiations, type-check failures while annotating subterms, and `validateBackendProgram` failures.
   - Do not silently skip modules, declarations, bindings, unsupported type applications, constructor/case shapes, or invalid IR; return a structured conversion error instead.

4. Convert checked metadata into backend module/data declarations.
   - `CheckedProgram` -> `BackendProgram`, preserving module order and `checkedProgramMain`.
   - `CheckedModule` -> `BackendModule`, preserving checked module names and all checked bindings.
   - `DataInfo`/`ConstructorInfo` -> `BackendData`/`BackendConstructor`, using stable runtime constructor names (`ctorRuntimeName`) and semantically consistent data/type names so validation sees the same names in metadata and expressions.
   - Convert source metadata types (`SrcType`) to `BackendType`; reject `STVarApp` and other shapes the #19 IR cannot represent.

5. Convert checked xMLF terms with explicit carried types.
   - Build a conversion context from all checked global binding types plus constructor/data metadata.
   - Recursively call `typeCheckWithEnv` as needed to compute each subexpression result type, then emit typed `BackendExpr` nodes.
   - Map supported terms directly: `EVar`, `ELit`, `ELam`, `EApp`, `ELet`, `ETyAbs`, `ERoll`, and `EUnroll`.
   - For `ETyInst`, support the app-like forms that correspond to backend type application (`InstApp` and the existing canonical app-like sequence forms); reject other instantiation witnesses explicitly unless there is a mechanically justified no-op case such as `InstId`.
   - Preserve let/lambda/type-binder environments so local variable references validate against the same types used by the checked term.

6. Recover explicit backend data operations where the checked term shape provides enough evidence.
   - Detect saturated constructor applications headed by a known constructor runtime name and emit `BackendConstruct` with converted argument expressions.
   - Detect the deferred-case rewrite shape produced by `.mlfp` finalization, namely an unrolled scrutinee type-instantiated with a result type and applied to constructor handlers, and emit `BackendCase` with constructor-pattern alternatives in data constructor order.
   - Accept only handler shapes whose lambda binders match the constructor arity; reject unclear or unsupported case encodings instead of falling back to opaque applications.
   - Leave partial constructor values as ordinary typed function expressions when they are valid checked xMLF terms.

7. Validate the produced IR before returning it.
   - Run `validateBackendProgram` at the end of `convertCheckedProgram`.
   - Wrap any `BackendValidationError` in `BackendConversionError` so callers never receive invalid backend IR.

8. Add focused tests.
   - Add `test/BackendConvertSpec.hs`; wire it into `mlf2.cabal` and `test/Main.hs`.
   - Cover a simple checked function/application program and assert conversion succeeds, preserves binding types, and validates.
   - Cover an ADT constructor/case program from the current `.mlfp` surface and assert conversion emits backend data metadata plus explicit `BackendConstruct`/`BackendCase` nodes where supported.
   - Cover unsupported construct reporting with a direct checked-term or checked-program fixture that triggers an unsupported instantiation/type shape and assert a `BackendConversionError`, not a dropped declaration or invalid IR.
   - Add a validation-propagation test for a deliberately invalid converted shape if the module exposes a suitable helper.

9. Update durable docs for the new boundary.
   - Update `docs/architecture.md` to mention `MLF.Backend.Convert` as the checked-program-to-typed-IR conversion owner and keep LLVM/text lowering reserved for #21.
   - Add a concise `CHANGELOG.md` entry because this is meaningful backend pipeline progress.

10. Validate and publish through PR #46.
   - Run a narrow test first, e.g. the new `BackendConvertSpec` plus existing `BackendIRSpec`.
   - Run the full required gate: `cabal build all && cabal test`.
   - Before committing, inspect `git status --short` and relevant diffs; stage only issue-related files.
   - Commit as `codex-watcher <codex-watcher@users.noreply.github.com>` with an imperative message.
   - Run `gh auth status` and `gh auth setup-git`, then push `codex/issue-20` to the existing PR branch. Do not create another PR.

---


src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,638 @@
...
+                    backendScrutinee = scrutineeExpr,
+                    backendAlternatives = alternatives
+                  }
+            )
+
+caseHead :: ElabTerm -> Maybe (ElabTerm, ElabType)
+caseHead term =
+  case term of
+    ETyInst (EUnroll scrutinee) inst
+      | Just scrutineeTy <- appLikeInstantiationType inst -> Just (scrutinee, scrutineeTy)
```
soulomoon: This `InstApp` payload is the case result type, not the scrutinee type. Passing it down as the expected scrutinee type makes supported cases fail when the result differs from the ADT: for example, a checked `case Succ Zero of { Zero -> 0; Succ n -> 1 }` currently returns `BackendUnsupportedCaseShape "no backend data matches scrutinee type ... Int"` because the constructor scrutinee is converted as `Int`. The converter should infer the scrutinee term's actual type (or otherwise use the type of `scrutinee`) before calling `requireCaseData`.

soulomoon (commented):
Found one conversion correctness issue.

src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,762 @@
...
+        BackendApp
+          { backendExprType = resultTy,
+            backendFunction = funExpr,
+            backendArgument = argExpr
+          }
+    ELet name scheme rhs body -> do
+      let schemeTy = schemeToType scheme
+          envForRhs = extendTermEnv name schemeTy env
+      bindingTy <- convertElabType schemeTy
+      rhsExpr <- convertTermExpected context envForRhs (Just bindingTy) rhs
```
soulomoon: `BackendLet` is validated as a non-recursive let, but this conversion type-checks and converts the RHS with the let name already in scope. A checked recursive local let therefore produces IR whose RHS can contain `BackendVar name`, and final validation rejects it as an unknown variable. For example, converting `test/programs/unified/authoritative-recursive-let.mlfp` at this head returns `BackendValidationFailed (BackendUnknownVariable "$peel#0")`. The converter should either represent recursive lets in the backend IR or reject recursive `ELet` explicitly before emitting invalid IR.

soulomoon (commented):
Found one remaining conversion correctness issue.

src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,789 @@
...
+convertTermExpected :: ConvertContext -> Env -> Maybe BackendType -> ElabTerm -> Either BackendConversionError BackendExpr
+convertTermExpected context env mbExpectedTy term =
+  case mbExpectedTy of
+    Just resultTy ->
+      convertSpecialTerm context env term resultTy
+        >>= \case
+          Just expr -> Right expr
+          Nothing -> do
+            inferredTy <- inferBackendType env term
+            convertOrdinaryTerm context env term inferredTy
```
soulomoon: When an expected type is available, this fallback still throws it away and converts ordinary terms with the inferred type. That loses source annotations/newtype-style rolls that the checked binding type is supposed to preserve. At this head, converting `test/programs/recursive-adt/abstract-module-use.mlfp` builds `Core__succ` with declared type `Nat -> Nat`, but the lambda is carried as `Nat -> forall ...`; the body constructor then validates as `Core__Succ` returning the unrolled `forall` type and `convertCheckedProgram` fails with `BackendConstructorResultMismatch`. The ordinary fallback should preserve/check the supplied expected type instead of always re-infering the result type.

soulomoon (commented):
Found one additional conversion correctness issue at this head.

src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,795 @@
...
+    BackendModule
+      { backendModuleName = checkedModuleName checkedModule,
+        backendModuleData = dataDecls,
+        backendModuleBindings = bindings
+      }
+
+convertCheckedBinding :: ConvertContext -> Env -> CheckedBinding -> Either BackendConversionError BackendBinding
+convertCheckedBinding context env binding = do
+  bindingTy <- convertElabType (checkedBindingType binding)
+  expr <- convertTermExpected context env (Just bindingTy) (checkedBindingTerm binding)
```
soulomoon: Converting every checked binding body directly with the exported binding type still breaks current checked data examples whose constructor bindings carry extra Church-encoding type abstractions. At this head, `convertCheckedProgram` on `test/programs/recursive-adt/recursive-gadt.mlfp` and `recursive-existential.mlfp` fails with `BackendValidationFailed (BackendTypeAbsTypeMismatch ...)` while converting generated constructor bindings like `RecursiveGadt__DoneNat`/`RecursiveGadt__Step`. That means the converter emits invalid IR for checked GADT/existential data paths instead of either representing those constructor bindings correctly or rejecting that unsupported shape explicitly before validation, so the issue’s current-example data-path/explicit unsupported-reporting criteria are not satisfied.

soulomoon (commented):
Found one remaining task-completion issue against the checked-program conversion scope.

src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,1042 @@
...
+            backendLetBody = bodyExpr
+          }
+    ETyAbs name mbBound body -> do
+      mbBackendBound <- traverse (convertElabType . tyToElab) mbBound
+      let boundTy = maybe TBottom tyToElab mbBound
+          bodyExpected =
+            case resultTy of
+              BTForall _ _ bodyTy -> Just bodyTy
+              _ -> Nothing
+      bodyExpr <- convertTermExpected context (extendTypeEnv name boundTy env) bodyExpected body
```
soulomoon, line 597: This descends into the body using `bodyExpected` from `resultTy`, but that expected forall body has not been renamed to the actual `ETyAbs` binder `name`. When the checked binding's source type is alpha-equivalent but uses a different binder name than the elaborated term, the body is converted under `BTVar "a"` while lambda parameters are still entered in the environment as `TVar "a1"`, so validation rejects the IR. A current repro is `test/programs/unified/first-class-polymorphism.mlfp`, where `convertCheckedProgram` returns `BackendValidationFailed (BackendVariableTypeMismatch "$x#0" (BTVar "a1") (BTVar "a"))`. Rename/substitute the expected forall body to the term binder before recursing, or keep the inferred binder names, so checked first-class polymorphic definitions convert to valid backend IR.

soulomoon (commented):
Found one remaining conversion correctness issue.

src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,1119 @@
...
+                  initialSubstitution <- constructorTypeApplicationSubstitution constructorMeta headTypeArgs
+                  resultSubstitution <-
+                    case matchBackendTypeParameters parameters initialSubstitution (backendConstructorResult constructor) resultTy of
+                      Just substitution -> Right substitution
+                      Nothing ->
+                        Left
+                          ( BackendUnsupportedCaseShape
+                              ("constructor result type does not match expected result for `" ++ backendConstructorName constructor ++ "`")
+                          )
+                  substitution <-
```
soulomoon, line 717: This new pre-match rejects valid nested constructor arguments before they can be converted with the expected field type. For `test/programs/recursive-adt/typeclass-integration.mlfp`, `TypeclassIntegration__Succ`'s checked binding has the Church-encoded result type `Nat -> forall a1. ...`, so `inferBackendType` for the nested `Succ Zero` argument returns that unrolled forall body instead of the metadata field type `Nat`. Conversion now fails with `BackendUnsupportedCaseShape "constructor argument type does not match case scrutinee constructor `TypeclassIntegration__Succ`"`, whereas the previous code converted the argument under the expected `Nat` and recovered `BackendConstruct`. The constructor-field matching should not hard-reject before expected-type conversion has a chance to canonicalize nested constructor applications.

soulomoon (commented):
Found one conversion regression in the current-example corpus.

soulomoon (commented):
Clean review at 1fb785e. I re-read the PR plan, linked issue, and current review history; the prior actionable threads are resolved/outdated, and the current head satisfies the conversion boundary plan and issue acceptance criteria based on the inspected code and prior validation.

Reviewed commit: `1fb785e732bd2e59956ccfc193f1a5786288f079`

src/MLF/Backend/Convert.hs
```diff
@@ -0,0 +1,1114 @@
...
+        >>= Right . dmBackend
+
+convertConstructorInfo :: ElaborateScope -> ConstructorInfo -> Either BackendConversionError BackendConstructor
+convertConstructorInfo scope info = do
+  fields <- mapM (convertLoweredSourceType scope) (ctorArgs info)
+  resultTy <- convertLoweredSourceType scope (ctorResult info)
+  Right
+    BackendConstructor
+      { backendConstructorName = ctorRuntimeName info,
+        backendConstructorForalls = map fst (ctorForalls info),
```
soulomoon, line 464: `ctorForalls` carries optional bounds, but the backend metadata keeps only the binder names here. That means a checked constructor like `Pack : forall (a >= Int). a -> Pack` converts to metadata with `backendConstructorForalls = ["a"]` and `backendConstructorFields = [BTVar "a"]`; the bound is only present inside the Church-encoded result type, not in the constructor parameter metadata used for argument validation. I verified that a `BackendConstruct Pack` with a `Bool` argument validates once this metadata is built, so the backend IR no longer preserves/enforces the constructor-local bound. Please carry `Maybe BackendType` bounds in `BackendConstructor` and use them when matching constructor arguments/pattern fields.

soulomoon (commented):
Found one post-merge backend IR type-preservation gap.

PR #47, opened 2026-04-27
Implement #20

### LINKED ISSUE #20: Convert checked .mlfp programs to typed backend IR

soulomoon:
Closes #20.

## Implementation Plan

---
issue_number: 20
pr_number: 47
branch: codex/issue-20-2
---

## Implementation Plan

1. Start from the prepared PR branch.
   - Work in `/workspace/artifacts/issue-workdirs/soulomoon_mlf2__issue20` on `codex/issue-20-2` for PR #47.
   - Re-check `git status --short --branch` before editing. The previous local `issue-plan.md` mentions PR #46 / `codex/issue-20`; treat that as stale.

2. Preserve constructor-local `forall` bounds in backend metadata.
   - In `src/MLF/Backend/IR.hs`, add an exported backend type-binder representation, e.g. `BackendTypeBinder { backendTypeBinderName :: String, backendTypeBinderBound :: Maybe BackendType }`.
   - Change `BackendConstructor.backendConstructorForalls` from `[String]` to `[BackendTypeBinder]` so constructor-local bounds are part of the typed IR.
   - Add small internal helpers for binder names and parameter-bound maps rather than open-coding tuple/list logic.

3. Convert source constructor bounds instead of dropping them.
   - In `src/MLF/Backend/Convert.hs`, update `convertConstructorInfo` to convert each `(name, Maybe SrcType)` from `ctorForalls` into a backend binder, using the module-aware lowered type path for bounds.
   - Update all constructor parameter-name logic, including `constructorTypeParameters`, `constructorTypeApplicationParameterNames`, and `dataMatchesScrutinee`, to use binder names from the new metadata.

4. Enforce bounds while matching constructor parameters.
   - Replace the current `Set String` parameter matcher in both `MLF.Backend.IR` and the converter-local matcher with a `Map String (Maybe BackendType)` or equivalent.
   - Data type parameters should remain unbounded; constructor-local binders should carry their converted bounds.
   - When inserting or reusing a substitution for a bounded parameter, require the actual type to match the substituted bound, following the existing backend type-application semantics: `Nothing`/`BTBottom` permits any type, `Just bound` requires `alphaEqBackendType` after applying known substitutions.
   - Preserve the existing alpha-equivalent repeated-substitution behavior.

5. Apply the bound-aware matcher everywhere constructor metadata is trusted.
   - Update constructor validation for `BackendConstruct` arguments.
   - Update constructor-pattern validation so bounded constructor-local parameters left unresolved by the scrutinee/result are instantiated from their bounds before pattern binder field types are exposed.
   - Update converter paths that infer constructor application results, recover explicit `BackendConstruct`, recover `BackendCase`, synthesize constructor bindings, and select matching data metadata.

6. Add regression tests in existing spec modules.
   - In `test/BackendIRSpec.hs`, add a bounded constructor metadata fixture such as `Pack : forall (a >= Int). a -> Pack`.
   - Assert that `BackendConstruct Pack [Int]` validates and `BackendConstruct Pack [Bool]` fails with a constructor argument mismatch using `Int` as the expected type.
   - Add a case-pattern check proving a bounded constructor field binder is typed as the bound, not as an unconstrained free `BTVar`.
   - In `test/BackendConvertSpec.hs`, add a checked `.mlfp` program with `Pack : forall (a >= Int). a -> Pack`; assert conversion preserves `Just intTy` in `backendConstructorForalls`, the converted program validates, and a manually corrupted Bool construction using the converted metadata is rejected.

7. Update durable docs for the changed IR contract.
   - In `docs/architecture.md`, mention that backend constructor metadata preserves constructor-local `forall` bounds and validation enforces them at construct/case boundaries.
   - Add a concise `CHANGELOG.md` entry because this fixes a meaningful backend IR type-preservation gap.

8. Validate before publish.
   - Run a narrow Hspec slice for backend specs, for example `cabal test mlf2-test --test-options='--match MLF.Backend'`.
   - Run the required full gate before completion: `cabal build all && cabal test`.
   - Before committing, inspect `git status --short` and relevant diffs; stage only issue-related files.
   - Commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, run `gh auth status` and `gh auth setup-git`, then push `codex/issue-20-2` to the existing PR #47. Do not create another PR or resolve PR review threads in this implementation turn.

---

### DIFF
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -1158,7 +1158,11 @@ matchBackendTypeParameters typeBounds parameterBounds =
               True
         Just (Just boundTy)
           | not (alphaEqBackendType boundTy BTBottom) ->
-              let expectedBound = substituteBackendTypes (Map.delete name substitution) boundTy
+              let dependencySubstitution =
+                    completeBackendParameterSubstitution
+                      (Map.delete name parameterBounds)
+                      (Map.delete name substitution)
+                  expectedBound = substituteBackendTypes dependencySubstitution boundTy
                in alphaEqBackendType actual expectedBound || actualTypeVariableBoundMatches actual expectedBound
         _ ->
           True
@@ -1176,11 +1180,40 @@ matchBackendTypeParameters typeBounds parameterBounds =
 
 completeBackendParameterSubstitution :: BackendParameterBounds -> Map String BackendType -> Map String BackendType
 completeBackendParameterSubstitution parameterBounds substitution0 =
-  foldl insertBoundDefault substitution0 (Map.toList parameterBounds)
+  resolveDefaultedBounds defaultedNames substitution1
   where
+    substitution1 =
+      foldl insertBoundDefault substitution0 (Map.toList parameterBounds)
+
+    defaultedNames =
+      Set.fromList
+        [ name
+          | (name, Just boundTy) <- Map.toList parameterBounds,
+            Map.notMember name substitution0,
+            not (alphaEqBackendType boundTy BTBottom)
+        ]
+
     insertBoundDefault substitution (name, Just boundTy)
       | Map.member name substitution = substitution
       | alphaEqBackendType boundTy BTBottom = substitution
       | otherwise = Map.insert name (substituteBackendTypes substitution boundTy) substitution
     insertBoundDefault substitution _ =
       substitution
+
+    resolveDefaultedBounds names =
+      go (Set.size names + Map.size parameterBounds + 1)
+      where
+        go remaining substitution
+          | remaining <= 0 = substitution
+          | substitution' == substitution = substitution
+          | otherwise = go (remaining - 1) substitution'
+          where
+            substitution' =
+              foldl resolveDefaultedBound substitution (Set.toList names)
+
+    resolveDefaultedBound substitution name =
+      case Map.lookup name substitution of
+        Just ty ->
+          Map.insert name (substituteBackendTypes (Map.delete name substitution) ty) substitution
+        Nothing ->
+          substitution
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -1163,20 +1163,31 @@ matchBackendTypeParameters typeBounds parameterBounds =
                       (Map.delete name parameterBounds)
                       (Map.delete name substitution)
                   expectedBound = substituteBackendTypes dependencySubstitution boundTy
-               in alphaEqBackendType actual expectedBound || actualTypeVariableBoundMatches actual expectedBound
+               in typeBoundDependenciesMatch actual expectedBound || actualTypeVariableBoundMatches actual expectedBound
         _ ->
           True
+
+    typeBoundDependenciesMatch actual expectedBound =
+      alphaEqBackendType
+        (resolveTypeBoundDependencies actual)
+        (resolveTypeBoundDependencies expectedBound)
 
     actualTypeVariableBoundMatches actual expectedBound =
       case actual of
         BTVar actualName ->
           case Map.lookup actualName typeBounds of
             Just (Just actualBound) ->
-              alphaEqBackendType actualBound expectedBound
+              typeBoundDependenciesMatch actualBound expectedBound
             _ ->
               False
         _ ->
           False
+
+    resolveTypeBoundDependencies =
+      substituteBackendTypes resolvedTypeBounds
+
+    resolvedTypeBounds =
+      completeBackendParameterSubstitution typeBounds Map.empty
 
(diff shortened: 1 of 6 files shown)


src/MLF/Backend/IR.hs
```diff
@@ -507,9 +518,27 @@ validateBackendVariable (Just context0) name actualTy =
     Nothing ->
       Left (BackendUnknownVariable name)
     Just expectedTy ->
-      unless (alphaEqBackendType actualTy expectedTy) $
+      unless (backendVariableTypeMatches (bvcTypeBounds context0) expectedTy actualTy) $
         Left (BackendVariableTypeMismatch name expectedTy actualTy)
 
+backendVariableTypeMatches :: Map.Map String (Maybe BackendType) -> BackendType -> BackendType -> Bool
+backendVariableTypeMatches typeBounds expectedTy actualTy =
+  alphaEqBackendType actualTy expectedTy
```
soulomoon: **P1: Match bounded variables recursively in carried types**

This only treats a bounded existential as interchangeable with its bound when the whole variable type is exactly `BTVar a`. Pattern fields can contain the unresolved constructor-local variable under another type constructor, e.g. `Wrap : forall (a >= Int). List a -> Wrap`; after the case-pattern change the binder is exposed as `List a` with `a >= Int`, but a valid handler body that uses it at `List Int` is rejected with `BackendVariableTypeMismatch`. The variable-type comparison needs to recurse through the type shape, or otherwise instantiate unresolved bounded variables inside compound field types while still preserving the re-pack case.

soulomoon (commented):
I found one correctness issue in the current head.

src/MLF/Backend/Convert.hs
```diff
@@ -1108,7 +1142,78 @@ matchBackendTypeParameters parameters =
...
+               in alphaEqBackendType actual expectedBound || actualTypeVariableBoundMatches actual expectedBound
+        _ ->
+          True
+
+    actualTypeVariableBoundMatches actual expectedBound =
+      case actual of
+        BTVar actualName ->
+          case Map.lookup actualName typeBounds of
+            Just (Just actualBound) ->
+              alphaEqBackendType actualBound expectedBound
```
soulomoon: This compares the actual variable's immediate bound against the constructor bound, so equivalent bounds that need the surrounding type-bound environment are still rejected. For example, with `Pack : forall (a >= List Int). a -> Pack`, a checked-like wrapper `forall (z >= Int). forall (b >= List z). b -> Pack` should allow `Pack x`, but `typeBounds` has `b -> List z` and `z -> Int`, and this line compares `List z` directly to `List Int`. The converter then drops the `a -> b` substitution, defaults `a` to `List Int`, and emits a backend program that fails validation. Please normalize the actual bound through `typeBounds` recursively before this comparison.

src/MLF/Backend/IR.hs
```diff
@@ -709,13 +812,79 @@ matchBackendTypeParameters parameters =
...
+               in alphaEqBackendType actual expectedBound || actualTypeVariableBoundMatches actual expectedBound
+        _ ->
+          True
+
+    actualTypeVariableBoundMatches actual expectedBound =
+      case actual of
+        BTVar actualName ->
+          case Map.lookup actualName typeBounds of
+            Just (Just actualBound) ->
+              alphaEqBackendType actualBound expectedBound
```
soulomoon: The validator has the same immediate-bound comparison as the converter, so backend IR that preserves an equivalent dependent bound is rejected. A constructor bound like `a >= List Int` should match an actual field variable `b` when the surrounding type bounds say `b >= List z` and `z >= Int`; today this compares `List z` to `List Int` without resolving `z`. Please resolve type-bound dependencies before checking alpha-equivalence here as well.

soulomoon (commented):
I found a remaining bounded-constructor correctness issue in the current head.

soulomoon (commented):
Clean review. I re-read PR #47 and issue #20, checked the current head at 6ace098d11be83863bcbee07ce1a2d2da6000797, and found no actionable gaps against the implementation plan or acceptance criteria. Validation passed with `cabal test mlf2-test --test-options='--match MLF.Backend'`, `cabal build all && cabal test`, and `git diff --check`.

Reviewed commit: `6ace098d11be83863bcbee07ce1a2d2da6000797`

PR #77, opened 2026-04-28
Implement #68

### LINKED ISSUE #68: Extract shared ProgramSpec-to-LLVM parity matrix
## Scope
Extract the interpreter-success surface from `ProgramSpec` into a shared parity matrix that can be run by both the interpreter path and the LLVM backend path. Include `fixturePaths`, `unifiedFixtureExpectations`, `ExpectRunValue` cases from the eMLF surface and boundary matrices, and any standalone `runProgram` success checks that are part of the source-program runtime surface.

## Acceptance Criteria
- Interpreter assertions keep their existing expected values.
- LLVM parity assertions emit LLVM for every interpreter-success case and validate it with `llvm-as`.
- Representative cases run `llc -filetype=obj -o /dev/null` without making the whole suite unreasonably slow.
- Temporary `ExpectLLVMUnsupported` classifications are explicit, named, and mechanically counted so no success case is silently omitted.

## Dependencies / Blockers
- No dependency on other sub-issues.
- This should land before backend feature-completion work so siblings share one canonical case list.

This sub-issue should be test-harness and matrix organization only. It may expose unsupported LLVM cases, but it should not implement backend semantics that belong to the feature-focused sibling sub-issues.

soulomoon:
Closes #68.

## Implementation Plan

---
issue_number: 68
pr_number: 77
branch: codex/issue-68
---

1. Confirm implementation starts from the clean `codex/issue-68` branch and that PR #77 still targets `master` from `codex/issue-68`.
2. Extract the reusable ProgramSpec runtime matrix into a test-support module, preferably `test/Parity/ProgramMatrix.hs` with module name `Parity.ProgramMatrix`. Move/export `fixturePaths`, `unifiedFixtureExpectations`, `ProgramMatrixSource`, `ProgramMatrixExpectation`, `ProgramMatrixCase`, `emlfSurfaceParityMatrix`, and `emlfBoundaryMatrix` from `test/ProgramSpec.hs`.
3. In the shared matrix module, add an aggregate interpreter-success list for LLVM parity, built from all `ExpectRunValue` cases in the eMLF surface and boundary matrices, all `fixturePaths` with expected `true`, all `unifiedFixtureExpectations`, and the fixed-value standalone `runProgram` success cases in `ProgramSpec` that are part of source-program runtime behavior.
4. Update `test/ProgramSpec.hs` to import the shared matrix and keep the existing interpreter assertions and expected values. Leave non-fixed-value runtime assertions and diagnostic/check-failure assertions in `ProgramSpec` unless they can be represented without weakening their current checks.
5. Update `test/BackendLLVMSpec.hs` with a ProgramSpec-to-LLVM parity section. For every aggregate interpreter-success case, load the shared source, check it, render LLVM with `renderCheckedProgramLLVM`, and validate supported cases with `validateLLVMAssembly`.
6. Add an explicit named LLVM parity expectation/classification layer in `BackendLLVMSpec`, including `ExpectLLVMUnsupported` entries keyed by case name for cases the backend still cannot lower. Assert the unsupported count and key coverage mechanically so no interpreter-success case is silently omitted and stale unsupported entries fail once support lands.
7. Add a small explicit representative set of supported parity case names that also runs `validateLLVMObjectCode` with `llc -filetype=obj -o /dev/null`, keeping this list small to avoid slowing the full suite.
8. Add `Parity.ProgramMatrix` to the `mlf2.cabal` test-suite `other-modules`. Do not update `test/Main.hs` unless a new spec module is introduced; a test-support module only needs the Cabal stanza.
9. Validate incrementally with a focused Hspec match for `MLF.Program` and `MLF.Backend.LLVM`, then run the required full gate `cabal build all && cabal test`. If LLVM tools are unavailable, rely on the existing `pendingWith` behavior and report that limitation.
10. Before publishing in the implementation turn, inspect `git status --short` and relevant diffs, stage only issue #68 files, commit with the configured watcher identity, run `gh auth status` and `gh auth setup-git`, then push `codex/issue-68` to update existing PR #77 without creating another PR.

---


soulomoon (commented):
LGTM. The PR body and issue #68 are satisfied at 863beacb3b0a0a922a1d958ca1389c10c07947b5. I validated the focused Program/LLVM specs and the full gate with `cabal build all && cabal test`.

Reviewed commit: `863beacb3b0a0a922a1d958ca1389c10c07947b5`

PR #82, opened 2026-04-29
Implement #72

### LINKED ISSUE #72: Support first-class polymorphism in LLVM parity
## Scope
Make LLVM parity pass for ProgramSpec interpreter-success cases with first-class polymorphism and forall-bearing runtime boundaries: rank-2 arguments, polymorphic local/top-level values, constructor fields carrying `forall`, pattern-bound polymorphic variables, and the `first-class-polymorphism.mlfp` fixture.

## Acceptance Criteria
- First-class polymorphic values that the interpreter runs lower to valid LLVM without escaping `BTForall`/function values into unsupported runtime positions.
- Constructor-bound polymorphic values used by successful ProgramSpec cases can be represented, specialized, erased, or otherwise compiled consistently with interpreter behavior.
- Existing backend rejection tests for unsupported partial applications or escaping closures are updated only when the implemented representation genuinely supports them.
- LLVM output validates with `llvm-as`; at least one first-class-polymorphism fixture receives `llc` smoke coverage.

## Dependencies / Blockers
- Depends on `Extract shared ProgramSpec-to-LLVM parity matrix` and the core first-order parity slice.
- May block final removal of `ExpectLLVMUnsupported` classifications for rank-2 and constructor-polymorphism cases.

Coordinate with recursive ADT and typeclass slices only at shared backend IR/type representation boundaries. Avoid taking over deriving, instance resolution, or qualified import identity work unless required to compile the polymorphic case itself.

soulomoon:
Closes #72.

## Implementation Plan

---
issue_number: 72
pr_number: 82
branch: codex/issue-72
---

## Implementation Plan

1. Keep the change scoped to the backend LLVM lowering path. The converter already preserves the needed IR shapes (`BackendTyAbs`, `BackendTyApp`, `BackendConstruct`, `BackendCase`); current failures come from `MLF.Backend.LLVM.Lower` trying to lower `BTForall`/function values as runtime LLVM operands.

2. In `src/MLF/Backend/LLVM/Lower.hs`, add a static function/value binding path that reuses `ExprEnv.eeLocalFunctions` for first-class polymorphic or function-typed values. Implement helpers to classify runtime-lowerable types, turn global/local/direct lambda/type-abs/alias expressions into `LocalFunction`s, and bind call arguments as either LLVM operands or static functions. Keep partial applications and escaping lambdas unsupported unless they are immediately called through this static path.

3. Adjust call lowering so global functions with unlowerable parameters, such as `usePoly : (forall a. a -> a) -> Bool`, are statically inlined at call sites instead of emitted as LLVM functions with `BTForall` parameters. Keep ordinary first-order global calls as direct LLVM calls, and keep existing polymorphic specialization behavior for type-applied globals.

4. Add an immediate `case (Construct ...) of ...` lowering path. When the selected constructor alternative binds a used field, bind lowerable fields as LLVM values and `forall`/function fields as static functions. Leave the normal heap allocation path unchanged, so constructor fields carrying polymorphic or function values still fail closed when they escape instead of being immediately destructed.

5. Update `test/BackendLLVMSpec.hs` with focused regression coverage for top-level `usePoly id`, local `usePoly id`, immediate `PolyBox` cases with polymorphic fields, and unchanged rejection for escaping function fields/partial applications. Then remove the now-supported issue-owned entries from `unsupportedLLVMParityCases`, reduce `expectedLLVMUnsupportedParityCount` from 48 to 42, and add `unified fixture: test/programs/unified/first-class-polymorphism.mlfp` to `llvmObjectCodeParityCases` for `llc` smoke coverage.

6. Add a concise `CHANGELOG.md` Unreleased entry describing LLVM support for first-class polymorphism through static specialization/erasure at non-escaping runtime boundaries.

7. Validate with focused backend tests first, then the full gate before publish: `cabal test mlf2-test --test-show-details=direct --test-options='--match "MLF.Backend.LLVM"'`, confirm the first-class-polymorphism fixture validates through LLVM assembly/object smoke coverage, then run `cabal build all && cabal test`.

---

### DIFF
--- a/src/MLF/Backend/LLVM/Lower.hs
+++ b/src/MLF/Backend/LLVM/Lower.hs
@@ -591,12 +591,10 @@ collectSpecializationRequestsWithBound base substitution bound expr =
             (BackendVar _ name, typeArgs)
               | Set.notMember name bound,
                 Just binding <- Map.lookup name (pbBindings base),
-                not (null (ffTypeBinders (biForm binding))),
-                null (ffParams (biForm binding)) ->
+                not (null (ffTypeBinders (biForm binding))) ->
                   let typeArgs' = map (substituteBackendTypes substitution) typeArgs
                    in case instantiateFunctionFormWithTypeArgs ("specialization request " ++ name) (biForm binding) typeArgs' [] of
-                        Right (resolvedTypeArgs, form)
-                          | null (ffParams form) -> [SpecRequest name resolvedTypeArgs]
+                        Right (resolvedTypeArgs, _) -> [SpecRequest name resolvedTypeArgs]
                         _ -> []
             (fun, typeArgs) ->
               collectAdministrativeTypeAppRequests fun typeArgs
@@ -1580,7 +1578,8 @@ lowerImmediateConstructCase env exprEnv context resultTy constructorName args fi
 
     bindImmediateAlternativePattern (BackendAlternative pattern0 body) =
       case pattern0 of
-        BackendDefaultPattern ->
+        BackendDefaultPattern -> do
+          zipWithM_ evaluateUnusedField fieldTys args
           pure exprEnv
         BackendConstructorPattern name binders -> do
           unless (name == constructorName) $
@@ -1611,6 +1610,17 @@ lowerImmediateConstructCase env exprEnv context resultTy constructorName args fi
             else pure acc
       | otherwise = do
           liftEither (BackendLLVMUnsupportedType ("field " ++ show binder ++ " at " ++ context) fieldTy)
+
+    evaluateUnusedField fieldTy arg
+      | backendTypeRequiresStaticSpecialization fieldTy = do
+          _ <- lowerStaticFunctionArgument env exprEnv context "_" fieldTy arg
+          pure ()
+      | backendTypeHasRuntimeRepresentation env fieldTy = do
+          value <- lowerExpr env exprEnv context arg
+          expectedTy <- lowerBackendTypeM env context fieldTy
+          requireLLVMType context constructorName expectedTy value
+      | otherwise =
+          liftEither (BackendLLVMUnsupportedType ("field at " ++ context) fieldTy)
 
 constructorFieldTypesForScrutinee :: ProgramEnv -> String -> ConstructorRuntime -> BackendType -> LowerM [BackendType]
 constructorFieldTypesForScrutinee _ context constructorRuntime scrutineeTy =
--- a/src/MLF/Backend/LLVM/Lower.hs
+++ b/src/MLF/Backend/LLVM/Lower.hs
@@ -329,7 +329,7 @@ backendTypeRequiresStaticSpecialization :: BackendType -> Bool
 backendTypeRequiresStaticSpecialization =
   \case
     BTVar {} -> False
-    BTArrow {} -> False
+    BTArrow {} -> True
     BTBase {} -> False
     BTCon {} -> False
     BTForall {} -> True
@@ -1597,16 +1597,18 @@ lowerImmediateConstructCase env exprEnv context resultTy constructorName args fi
             (zip3 binders fieldTys args)
 
     bindUsedField usedBinders acc (binder, fieldTy, arg)
-      | Set.notMember binder usedBinders =
-          pure acc
       | backendTypeRequiresStaticSpecialization fieldTy = do
           localFunction <- lowerStaticFunctionArgument env exprEnv context binder fieldTy arg
-          pure acc {eeLocalFunctions = Map.insert binder localFunction (eeLocalFunctions acc)}
+          if Set.member binder usedBinders
+            then pure acc {eeLocalFunctions = Map.insert binder localFunction (eeLocalFunctions acc)}
+            else pure acc
       | backendTypeHasRuntimeRepresentation env fieldTy = do
           value <- lowerExpr env exprEnv context arg
           expectedTy <- lowerBackendTypeM env context fieldTy
           requireLLVMType context constructorName expectedTy value
-          pure acc {eeValues = Map.insert binder value (eeValues acc)}
+          if Set.member binder usedBinders
+            then pure acc {eeValues = Map.insert binder value (eeValues acc)}
+            else pure acc
       | otherwise = do
           liftEither (BackendLLVMUnsupportedType ("field " ++ show binder ++ " at " ++ context) fieldTy)
 
--- a/src/MLF/Backend/LLVM/Lower.hs
+++ b/src/MLF/Backend/LLVM/Lower.hs
@@ -115,7 +115,8 @@ data FunctionState = FunctionState
     fsNextBlock :: Int,
     fsCurrentLabel :: String,
     fsCurrentInstructions :: [LLVMInstruction],
-    fsCompletedBlocks :: [LLVMBasicBlock]
+    fsCompletedBlocks :: [LLVMBasicBlock],
+    fsStaticGlobalStack :: [String]
   }
   deriving (Eq, Show)
 
@@ -138,7 +139,7 @@ lowerBackendProgram program = do
     (++)
       <$> traverse
         (lowerMonomorphicBinding env)
-        (filter (shouldEmitMonomorphicFunction env . biForm) (filter (null . ffTypeBinders . biForm) reachable))
+        (filter (shouldEmitReachableBinding env (backendProgramMain program)) (filter (null . ffTypeBinders . biForm) reachable))
       <*> traverse
         (lowerSpecialization env)
         (filter (shouldEmitMonomorphicFunction env . spForm) specializations)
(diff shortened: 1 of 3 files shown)


src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -136,8 +136,12 @@ lowerBackendProgram program = do
           }
   functions <-
     (++)
-      <$> traverse (lowerMonomorphicBinding env) (filter (null . ffTypeBinders . biForm) reachable)
-      <*> traverse (lowerSpecialization env) specializations
+      <$> traverse
+        (lowerMonomorphicBinding env)
+        (filter (shouldEmitMonomorphicFunction env . biForm) (filter (null . ffTypeBinders . biForm) reachable))
```
soulomoon: This filter can drop the program entrypoint itself when `main` has a static-only parameter, e.g. `main : (forall a. a -> a) -> Bool`. There is no call site where that binding can be inlined, so `emit-backend` currently succeeds with a module containing only declarations and no `main` definition instead of failing closed. The entrypoint needs to be excluded from this skip path or explicitly rejected when it cannot be emitted.

src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -1051,12 +1098,16 @@ lowerGlobalCall env exprEnv context name typeArgs args =
...
-      result <- emitAssign "call" resultTy (LLVMCall functionName [(lvLLVMType arg, lvOperand arg) | arg <- callArgs])
-      pure (LowerValue (ffReturnType form) resultTy result)
+      if functionFormHasRuntimeCallingConvention env form
+        then do
+          callArgs <- lowerRuntimeCallArguments env exprEnv context name form args
+          resultTy <- lowerBackendTypeM env context (ffReturnType form)
+          functionName <- globalFunctionName env context binding resolvedTypeArgs
+          result <- emitAssign "call" resultTy (LLVMCall functionName [(lvLLVMType arg, lvOperand arg) | arg <- callArgs])
+          pure (LowerValue (ffReturnType form) resultTy result)
+        else do
```
soulomoon: Unconditionally inlining static-argument globals here makes recursive static-bound definitions diverge during lowering. For example, `loop : (forall a. a -> a) -> Bool = \poly -> loop poly; main = loop id` never reaches an LLVM error or module because each inline of `loop` recursively lowers another `loop` call. This path needs recursion/cycle detection or a fail-closed fallback for recursive globals that cannot be emitted as LLVM functions.

soulomoon (commented):
Found two correctness issues in the LLVM static-lowering path.

soulomoon:
Review findings require rework at commit `ff4c6f7ef17bbd6019f03c0fc86388e4336b6b4b`.

Findings to address:
- `src/MLF/Backend/LLVM/Lower.hs:141` can filter out the entrypoint itself when `main` has a static-only parameter such as `(forall a. a -> a) -> Bool`, so `emit-backend` succeeds with no `main` definition instead of failing closed.
- `src/MLF/Backend/LLVM/Lower.hs:1108` unconditionally inlines static-argument globals, which makes recursive static-bound definitions diverge during LLVM lowering instead of producing a module or an unsupported error.


Submitted by Haskell PR review watcher as a non-approval review findings comment.

src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -288,6 +293,56 @@ collectForallsType =
...
+backendTypeHasRuntimeRepresentation env ty =
+  case lowerBackendType env "runtime representation check" ty of
+    Right _ -> True
+    Left _ -> False
+
+backendTypeRequiresStaticSpecialization :: BackendType -> Bool
+backendTypeRequiresStaticSpecialization =
+  \case
+    BTVar {} -> False
+    BTArrow {} -> False
```
soulomoon: The implementation plan calls out first-class polymorphic **or function-typed** values, but leaving `BTArrow` out of the static-specialization path means function arguments are still forced through runtime LLVM parameters. For example, `use : (Int -> Int) -> Int = \f -> f 1; main = use (\(x : Int) -> x)` still fails with `Unsupported backend LLVM type at parameter "$f#0" of Main__use` before it can use the static call path. This case is non-escaping and immediately called, so `BTArrow` parameters need the same static treatment (while still rejecting true escapes/partials).

soulomoon (commented):
Found one remaining task-completion gap in the static argument classification.

soulomoon:
Review findings require rework at commit `c0bf225d52eefe88166e5c1677001e43776d1093`.

Findings to address:
- `src/MLF/Backend/LLVM/Lower.hs:332` leaves `BTArrow` out of the static-specialization classification, so function-typed arguments are still forced through runtime LLVM parameters. A non-escaping immediately-called function argument such as `use : (Int -> Int) -> Int = \f -> f 1; main = use (\(x : Int) -> x)` still fails with `Unsupported backend LLVM type`, which does not satisfy the PR plan's function-typed static argument path.


Submitted by Haskell PR review watcher as a non-approval review findings comment.

src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -1081,12 +1139,104 @@ globalFunctionName env context binding typeArgs
...
+                expectedTy
+                ( LocalFunction
+                    { lfForm = biForm binding,
+                      lfCapturedEnv = emptyExprEnv
+                    }
+                )
+            Nothing ->
+              lowerDirectStaticFunctionArgument callEnv context paramName expectedTy arg
+    _ ->
+      lowerDirectStaticFunctionArgument callEnv context paramName expectedTy arg
```
soulomoon, line 2187: This fallback still mishandles type-applied global aliases used as static function arguments. For example, `id : forall a. a -> a; use : (Int -> Int) -> Int = \f -> f 1; main = use id` lowers the argument as a `LocalFunction` whose body calls `id @Int`, but no specialization for `id @Int` was collected, so lowering ends with `Internal backend LLVM error: missing specialization for Main__id`. Since global/type-applied aliases are part of the static function path, this needs either to instantiate the global binding directly as the local function or to register the required specialization before lowering.

soulomoon (commented):
Found one remaining static-argument lowering gap.

soulomoon:
Review findings require rework at commit `08baf45b93da4bd83e4d136ab05337c85d784094`.

Findings to address:
- `src/MLF/Backend/LLVM/Lower.hs:1211` still mishandles type-applied global aliases used as static function arguments. A program like `id : forall a. a -> a; use : (Int -> Int) -> Int = \f -> f 1; main = use id` reaches `Internal backend LLVM error: missing specialization for Main__id` because the static local function body calls `id @Int` without a collected specialization.


Submitted by Haskell PR review watcher as a non-approval review findings comment.

soulomoon:
Review findings require rework at commit `4029dd5158bfdf463b192cd4697a19809f93401f`.

Findings to address:
- pre-merge merge state is DIRTY: the PR branch is not mergeable with the latest base branch. Merge or rebase the latest base branch into the PR branch, resolve conflicts, rerun validation, and push the fix.


Submitted by Haskell PR review watcher as a non-approval review findings comment.

soulomoon (commented):
Clean review at 5b37d22b410d4bad70fa05fa5945f26b7d631d30. The workspace is at the target SHA, GitHub reports the PR mergeable, the changed LLVM lowering and regression tests match the PR plan and linked issue, and validation at this same SHA passed focused LLVM backend tests plus `cabal build all && cabal test`.

Reviewed commit: `5b37d22b410d4bad70fa05fa5945f26b7d631d30`

src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -2073,6 +2157,85 @@ globalFunctionName env context binding typeArgs
...
+    sourceTy = foldr BTArrow (ffReturnType form) (map snd (ffParams form))
+
+lowerDirectStaticFunctionArgument :: ExprEnv -> String -> String -> BackendType -> BackendExpr -> LowerM LocalFunction
+lowerDirectStaticFunctionArgument callEnv context paramName expectedTy arg =
+  requireStaticFunctionType
+    context
+    paramName
+    expectedTy
+    ( LocalFunction
+        { lfForm = functionFormFromExpected expectedTy arg,
```
soulomoon, line 2218: This fallback can accept an arbitrary function-typed expression as a zero-argument `LocalFunction` and then erase it when the static parameter is unused. For example, `use : (Int -> Int) -> Int = \f -> 0; main = use (add 1)` now emits `main = 0` even though `add 1` is a partial application and the plan says partial applications remain unsupported unless they are immediately called through the static path. Please restrict static arguments to direct lambda/type-abs/alias/global/local forms or otherwise force unsupported function-producing expressions to fail closed before erasure.

soulomoon (commented):
Found one remaining fail-closed gap in static function argument lowering.

PR #85, opened 2026-04-29
Implement #74

### LINKED ISSUE #74: Eliminate LLVM parity unsupported classifications and finish #63
## Scope
Complete the final #63 parity sweep after feature-focused sub-issues land. Remove temporary `ExpectLLVMUnsupported` classifications for interpreter-success cases, verify the shared matrix has no omissions, and keep `BackendLLVMSpec` focused on backend-specific regressions rather than broad source-program parity.

## Acceptance Criteria
- Every `ProgramSpec` interpreter runtime-success case in the shared matrix passes LLVM emission and `llvm-as` validation.
- No `ExpectLLVMUnsupported` remains for interpreter-success cases.
- Representative `llc -filetype=obj -o /dev/null` coverage is documented and balanced against suite runtime.
- Backend-specific regression tests remain in `BackendLLVMSpec`; broad parity coverage lives in the shared matrix runner.
- The parent issue #63 acceptance criteria can be checked directly from the test suite.

## Dependencies / Blockers
- Depends on all feature-focused parity sub-issues: core first-order, recursive ADT/recursive let, typeclass/deriving, first-class polymorphism, and higher-kinded/qualified constructor support.

This is an integration and cleanup slice. It should not introduce large new backend features; any remaining feature gap should be split back into a targeted sibling before closing this sub-issue.

soulomoon:
Closes #74.

## Implementation Plan

---
issue_number: 74
pr_number: 85
branch: codex/issue-74
---

# Implementation Plan

1. Start from a clean `codex/issue-74` worktree against `origin/HEAD`; reread the watcher-written `issue-plan.md`, verify PR #85 still targets `master` from `codex/issue-74`, and avoid any watcher state/runtime files.
2. Confirm the current remaining gap: `test/BackendLLVMSpec.hs` has one `ExpectLLVMUnsupported` entry for `standalone: does not decode typed non-data constructor fields through fallback ADT decoding`. Keep `programRuntimeSuccessCases = programSpecToLLVMParityCases`; do not remove this interpreter-success case from LLVM coverage.
3. In `src/MLF/Backend/LLVM/Lower.hs`, support stored first-order function constructor fields without broad closure conversion: derive constructor field types from metadata/result type before lowering `BackendConstruct` args; lower ordinary fields as today; lower first-order function fields to `LLVMPtr` via existing function-reference/static-function machinery; allow closed direct lambdas through a precollected private wrapper; continue rejecting captured closures or non-first-order function values.
4. Extend wrapper/reference collection so constructor-field function references and closed direct function expressions are emitted before use. Prefer generalizing the existing evidence-wrapper helpers into neutral static function wrappers rather than duplicating wrapper naming, keying, and closure checks.
5. Update heap case field loading so used first-order function fields load as `LLVMPtr` with the original backend function type, letting the existing indirect-call path handle calls from extracted fields.
6. Update `test/BackendLLVMSpec.hs`: replace the current `rejects function-typed constructor fields` expectation with positive coverage for a stored first-order function constructor field; add or keep a negative test for a captured/unsupported closure; remove `ExpectLLVMUnsupported`, `unsupportedLLVMParityCases`, and the fixed unsupported count so every `programSpecToLLVMParityCases` row must assemble.
7. Add the formerly unsupported standalone parity row to `llvmObjectCodeParityCases` if it keeps runtime acceptable, giving `llc -filetype=obj -o /dev/null` representative coverage for stored function-pointer constructor fields.
8. Update `CHANGELOG.md` and `implementation_notes.md` to record that the final LLVM parity unsupported classification was eliminated and that representative `llc` coverage includes the stored function-field case.
9. Validate with focused checks first: `rg -n 'ExpectLLVMUnsupported|unsupportedLLVMParityCases' test/BackendLLVMSpec.hs` should find no remaining unsupported classification; run the focused Backend LLVM parity matrix and the new function-field tests with `cabal test mlf2-test --test-show-details=direct --test-options=...`. Since `/usr/bin/llvm-as` and `/usr/bin/llc` are available, expect assembly and object-code checks to run rather than pend.
10. Run the required final gate `cabal build all && cabal test`. Before publishing, run `gh auth status` and `gh auth setup-git`, inspect `git status --short` and relevant diffs, stage only issue-related files, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, and push the existing `codex/issue-74` branch for PR #85 without creating a duplicate PR.

---

### DIFF
--- a/src/MLF/Backend/LLVM/Lower.hs
+++ b/src/MLF/Backend/LLVM/Lower.hs
@@ -56,6 +56,7 @@ data ProgramEnv = ProgramEnv
   { peBase :: ProgramBase,
     peSpecializations :: Map String Specialization,
     peEvidenceWrappers :: Map String EvidenceWrapper,
+    peFunctionWrappers :: Map String FunctionWrapper,
     peStringGlobals :: Map String String
   }
 
@@ -102,6 +103,14 @@ data EvidenceWrapper = EvidenceWrapper
   }
   deriving (Eq, Show)
 
+data FunctionWrapper = FunctionWrapper
+  { fwKey :: String,
+    fwFunctionName :: String,
+    fwExpectedType :: BackendType,
+    fwExpr :: BackendExpr
+  }
+  deriving (Eq, Show)
+
 data LowerValue = LowerValue
   { lvBackendType :: BackendType,
     lvLLVMType :: LLVMType,
@@ -140,13 +149,15 @@ lowerBackendProgram program = do
   reachable <- reachableBindings base (backendProgramMain program)
   specializations <- collectRequiredSpecializations base reachable
   let evidenceWrappers = collectEvidenceWrappers base reachable specializations
-      referencedFunctionNames = collectReferencedFunctionNames base reachable specializations evidenceWrappers
-      stringGlobals = assignStringGlobals (collectProgramStrings reachable specializations)
+      functionWrappers = collectFunctionWrappers base reachable specializations
+      referencedFunctionNames = collectReferencedFunctionNames base reachable specializations evidenceWrappers functionWrappers
+      stringGlobals = assignStringGlobals (collectProgramStrings reachable specializations evidenceWrappers functionWrappers)
       env =
         ProgramEnv
           { peBase = base,
             peSpecializations = Map.fromList [(specializationKey (spRequest spec), spec) | spec <- specializations],
             peEvidenceWrappers = Map.fromList [(ewKey wrapper, wrapper) | wrapper <- evidenceWrappers],
+            peFunctionWrappers = Map.fromList [(fwKey wrapper, wrapper) | wrapper <- functionWrappers],
             peStringGlobals = stringGlobals
           }
   functions <-
@@ -154,7 +165,8 @@ lowerBackendProgram program = do
       <$> sequence
         [ traverse (lowerMonomorphicBinding env) (filter (shouldLowerReachableBinding referencedFunctionNames) reachable),
           traverse (lowerSpecialization env) (filter (shouldLowerSpecialization referencedFunctionNames) specializations),
-          traverse (lowerEvidenceWrapper env) evidenceWrappers
+          traverse (lowerEvidenceWrapper env) evidenceWrappers,
+          traverse (lowerFunctionWrapper env) functionWrappers
         ]
   mainBinding <- requireBinding base (backendProgramMain program)
   when (not (null (ffTypeBinders (biForm mainBinding)))) $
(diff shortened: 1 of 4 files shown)


src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -1889,6 +2013,18 @@ lowerFunctionArgument env exprEnv context expectedTy arg =
     _ ->
       liftEither (BackendLLVMUnsupportedExpression context "unsupported function argument")
 
+lowerStoredFunctionArgument :: ProgramEnv -> ExprEnv -> String -> BackendType -> BackendExpr -> LowerM LowerValue
+lowerStoredFunctionArgument env exprEnv context expectedTy arg =
+  case collectTyApps arg of
+    (BackendVar _ name, typeArgs) ->
+      lowerFunctionReference env exprEnv context expectedTy name typeArgs
```
soulomoon: This `BackendVar` path also receives constructor-field binders introduced by the immediate-constructor case path. For `case Holder (\(x : Int) x) of { Holder f -> Holder f }`, `f` is stored in `eeLocalFunctions`; delegating to `lowerFunctionReference` reaches `lowerLocalFunctionReference`, which only accepts eta aliases, so LLVM emission fails with `unsupported function argument` even though the source program runs and the field lambda is closed. The direct closed-lambda wrapper needs to remain usable when an immediate constructor case re-stores the bound function field, or first-order function fields need to be bound as pointer values in this path.

soulomoon (commented):
Found one correctness issue in the new function-field lowering path.

soulomoon:
Review findings require rework at commit `06709a3e907d0ee12d64302b695373a2e43ddb49`.

Findings to address:
- `src/MLF/Backend/LLVM/Lower.hs` delegates stored function-field `BackendVar` arguments through `lowerFunctionReference`, but immediate constructor cases bind closed direct-lambda fields as `eeLocalFunctions`; re-storing such a pattern-bound field fails with `unsupported function argument` even though the source program runs. I left an inline comment on the new `lowerStoredFunctionArgument` path.

soulomoon (commented):
Found one remaining edge case in the stored function-field support.

src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -999,6 +1032,75 @@ evidenceWrapperKey :: BackendType -> BackendExpr -> String
...
+    wrapperRequests =
+      case expr of
+        BackendConstruct resultTy name args ->
+          case Map.lookup name (pbConstructors base) >>= \constructorRuntime -> constructorRuntimeFieldTypes constructorRuntime resultTy of
+            Just fieldTys ->
+              [ (fieldTy', arg')
+              | (fieldTy, arg) <- zip fieldTys args,
+                let fieldTy' = substituteBackendTypes substitution fieldTy,
+                let arg' = substituteExprTypes substitution arg,
+                isFirstOrderFunctionPointerType fieldTy',
```
soulomoon, line 1069: Skipping every simple function reference here leaves closed local function references without a wrapper. For example, `def main : Holder = let f : Int -> Int = \(x : Int) x in Holder f;` runs in the interpreter, but `emit-backend` still fails with `unsupported function argument "$f#..."`: `lowerStoredFunctionReference` falls back to `lfStoredReference`, yet no `FunctionWrapper` was collected because the constructor argument was just `f`. Top-level references can keep using direct globals, but closed local references need to be recognized here (or otherwise emitted through the static-function machinery) so stored first-order function fields are not limited to syntactic lambdas.

soulomoon:
Review findings require rework at commit `208a327aa05c1d627895fd2f59656da842ec526e`.

Findings to address:
- Stored function-field lowering still fails for closed local function references such as `let f : Int -> Int = \(x : Int) x in Holder f`; the interpreter accepts the program, but `emit-backend` reports `unsupported function argument` because the wrapper collector skips simple `BackendVar` constructor arguments and therefore never emits a `FunctionWrapper` for the local closed lambda behind `f`. I left an inline comment on the wrapper collection guard.

soulomoon (commented):
LGTM. The PR satisfies the implementation plan and linked issue criteria at the reviewed head; focused LLVM function-field checks and the full `cabal build all && cabal test` gate both pass.

Reviewed commit: `2971cc47f70d9fad72c0a7bc4fffc60f8cce58f3`

PR #103, opened 2026-04-29
Implement #9

### LINKED ISSUE #9: Support expected-type resolution for nullary overloaded .mlfp methods
Subtask split from #8.

## Goal

Support source-level first-order classes with nullary overloaded methods / associated values, so a `.mlfp` class like `Monoid` can be expressed and used ergonomically:

```mlfp
class Monoid a {
  mempty : a;
  append : a -> a -> a;
}

data Nat =
    Zero : Nat
  | Succ : Nat -> Nat;

instance Monoid Nat {
  mempty = Zero;
  append = \left \right left;
}

def main : Nat = append (mempty : Nat) Zero;
```

## Current Behavior

This is not a higher-kinded-type problem. `Monoid` is first-order in its class parameter, but `mempty : a` is a nullary overloaded method. The current `.mlfp` method-resolution path requires enough term arguments to identify the instance, so a bare overloaded method name remains ambiguous even when the surrounding source annotation appears to fix the expected type.

Local probe on 2026-04-21:

```mlfp
class Monoid a {
  mempty : a;
  append : a -> a -> a;
}

instance Monoid Nat {
  mempty = Zero;
  append = \(left : Nat) \(right : Nat) left;
}

def main : Nat = append (mempty : Nat) Zero;
```

Result:

```text
error: ambiguous overloaded method use `mempty`
hint: apply `mempty` to enough arguments or add an annotation that fixes the instance type
```

A first-order class with a unary method, e.g. `class Idd a { keep : a -> a; }`, works. So the missing behavior is specifically expected-type-driven resolution for nullary overloaded methods / associated values.

## Current Contract

`docs/mlfp-language-reference.md` currently documents this limitation: method calls lower to deferred obligations, partial overloaded method application is supported by eta-expanding missing arguments, and bare overloaded method names remain ambiguous and are rejected.

This issue should change that contract for the nullary-method case, if the design is accepted.

## Design Work Needed

Likely pieces:

- Preserve expected type information for overloaded method occurrences with zero term arguments.
- Teach deferred method resolution to use an explicit annotation or surrounding expected type as instance evidence when there are no value arguments.
- Keep ambiguity fail-closed: `mempty` without expected type evidence should still be rejected.
- Decide whether this is limited to class methods whose result is exactly the class parameter, or whether any method result containing the class parameter can use expected-type solving.
- Add positive `.mlfp` corpus coverage for `Monoid Nat` with `mempty : Nat` and use through `append`.
- Add negative coverage for ambiguous bare `mempty` with no expected type.
- Update `docs/mlfp-language-reference.md` and diagnostics once the supported behavior is finalized.

## Relationship To #8

#8 tracks higher-kinded/type-operator parameters such as `Functor f`, `Monad m`, and `Profunctor p`. This issue is narrower: first-order class methods that have no term arguments but can be resolved from expected type evidence.

soulomoon:
Closes #9.

## Implementation Plan

---
issue_number: 9
pr_number: 103
branch: codex/issue-9
---

## Implementation Plan

1. Keep the existing branch/PR setup intact: work on `codex/issue-9` for PR #103 and do not create another PR. Start implementation by reading this plan plus `git status --short --branch`.

2. Add expected-type lowering for bare nullary overloaded methods in `src/MLF/Frontend/Program/Elaborate.hs`:
   - Replace the immediate `ProgramAmbiguousMethodUse` for `EVar` / resolved `EVar` `OverloadedMethod` with a helper that only proceeds when `methodFullArity methodInfo == 0` and `mbExpected` is present.
   - Infer the class argument by matching the nullary method result type against the expected source type with `matchTypesInScope`; require the class parameter to be bound uniquely. Support any nullary result shape that determines the class parameter, not only result `a` exactly. Keep the old ambiguity error when there is no expected type or the expected type does not determine the class parameter.
   - Leave argument-driven overloaded method resolution unchanged for non-nullary methods and partial applications.

3. Preserve the expected type for finalization instead of relying on value arguments:
   - Extend `DeferredMethodCall` in `src/MLF/Frontend/Program/Types.hs` with an optional expected/result `TypeView` or equivalent source-type field for nullary occurrences.
   - Add or adapt a zero-argument method deferral helper so the placeholder external type is the lowered expected type and the deferred obligation records the expected source view. Existing nonzero method deferrals should record `Nothing` and keep their current placeholder behavior.

4. Teach `src/MLF/Frontend/Program/Finalize.hs` to rewrite zero-argument deferred methods:
   - In `resolveDeferredMethods`, detect a bare deferred method placeholder (including type-instantiated heads, as constructor finalization already does) when `deferredMethodArgCount == 0` and there are no term arguments.
   - Use the recorded expected/result type to recompute the class argument view, then reuse the existing instance/evidence machinery: `resolveMethodInstanceInfoByTypeView`, `concreteMethodValue`, method-local constraint evidence resolution, and `instantiateMethodValue`.
   - Fail closed with `ProgramAmbiguousMethodUse` if expected-type matching does not determine the class parameter, and with `ProgramNoMatchingInstance` if it does determine a type but no coherent instance exists.

5. Add coverage:
   - Add a `.mlfp` corpus fixture under `test/programs/unified/`, e.g. `authoritative-nullary-overloaded-method.mlfp`, with `Monoid Nat`, `mempty : a`, `append : a -> a -> a`, and `def main : Nat = append (mempty : Nat) Zero;`. Wire it into `test/Parity/ProgramMatrix.hs` with expected runtime value `Zero`.
   - Add inline matrix cases for `def main : Nat = mempty`, for expected-type propagation through an overloaded method argument such as `append mempty Zero` if the implementation supports it through existing expected-argument flow, and for a parameterized nullary result such as `defaultBox : Box a` resolving from expected `Box Nat`.
   - Keep or add a negative bare use case with no expected evidence, such as `let x = mempty in ...`, asserting `ProgramAmbiguousMethodUse`. Existing `eq` bare-method coverage should remain unchanged.

6. Update documentation and diagnostics:
   - Change `docs/mlfp-language-reference.md` so it no longer says all bare overloaded method names are rejected. Document that nullary overloaded methods / associated values resolve from an explicit or propagated expected source type when the method result determines the class parameter; otherwise they remain ambiguous.
   - Refine the `ProgramAmbiguousMethodUse` hint in `Program.Types` to mention expected-type annotations for nullary methods and sufficient term arguments for ordinary methods.
   - Add a concise `CHANGELOG.md` entry, and update `implementation_notes.md` if the repository’s recent-note style expects feature-level `.mlfp` behavior changes there.

7. Validate before publishing:
   - Run a focused test slice first, e.g. `cabal test mlf2-test --test-show-details=direct --test-options='--match "nullary overloaded"'` or the exact new matrix-case names.
   - Run `cabal run mlf2 -- run-program test/programs/unified/authoritative-nullary-overloaded-method.mlfp` and confirm it prints the expected value.
   - Finish with the repo gate required for behavior changes: `cabal build all && cabal test`.

8. Publish handoff after validation succeeds:
   - Run `gh auth status` and `gh auth setup-git`.
   - Inspect `git status --short` and relevant diffs; stage only issue-related source, test, fixture, and doc files.
   - Commit as `codex-watcher <codex-watcher@users.noreply.github.com>` with an imperative message such as `Support nullary overloaded method resolution`.
   - Push `codex/issue-9` to origin for existing PR #103, then return completion for watcher handoff.

---

### DIFF
--- a/src/MLF/Frontend/Program/Finalize.hs
+++ b/src/MLF/Frontend/Program/Finalize.hs
@@ -36,6 +36,7 @@ import MLF.Frontend.Program.Elaborate
     elaborateScopeDataTypes,
     elaborateScopeRuntimeTypes,
     inferClassArgument,
+    lookupEvidenceMethodByClass,
     lowerType,
     lowerTypeView,
     matchTypes,
@@ -49,6 +50,7 @@ import MLF.Frontend.Program.Types
     ConstructorInfo (..),
     ConstructorShape (..),
     DataInfo (..),
+    DeferredMethodEvidence (..),
     DeferredCaseCall (..),
     DeferredBindingMode (..),
     DeferredConstructorCall (..),
@@ -66,6 +68,7 @@ import MLF.Frontend.Program.Types
     constructorOwnerShapes,
     constructorShapeFromInfo,
     freeTypeVarsTypeView,
+    methodInfoOwnerClassSymbolIdentity,
     SymbolIdentity,
     splitArrows,
     splitForalls,
@@ -969,19 +972,37 @@ resolveDeferredMethods scope deferredMethods = go
         case inferNullaryMethodClassArgument methodInfo expectedView of
           Just view -> Right view
           Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
-      (instanceInfo, subst) <- resolveMethodInstanceInfoByTypeView scope methodInfo classArgView
-      methodValue <- concreteMethodValue instanceInfo methodInfo
-      methodSubst <-
-        case inferNullaryMethodSubst methodInfo classArgView subst expectedView of
-          Just subst' -> Right subst'
-          Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
-      let eagerConstraints =
-            filter
-              constraintGround
-              (map (applyConstraintInfoSubst methodSubst) (methodValueConstraints methodValue))
-      evidenceArgs <- resolveConstraintEvidenceTerms scope Set.empty eagerConstraints
-      methodHead <- instantiateMethodValue scope methodSubst methodValue
-      Right (foldl X.EApp methodHead evidenceArgs)
+      case lookupNullaryEvidence deferred methodInfo classArgView of
+        Just runtimeName -> Right (X.EVar runtimeName)
+        Nothing -> do
+          (instanceInfo, subst) <- resolveMethodInstanceInfoByTypeView scope methodInfo classArgView
+          methodValue <- concreteMethodValue instanceInfo methodInfo
+          methodSubst <-
+            case inferNullaryMethodSubst methodInfo classArgView subst expectedView of
+              Just subst' -> Right subst'
+              Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+          let eagerConstraints =
+                filter
+                  constraintGround
+                  (map (applyConstraintInfoSubst methodSubst) (methodValueConstraints methodValue))
+          evidenceArgs <- resolveConstraintEvidenceTerms scope Set.empty eagerConstraints
+          methodHead <- instantiateMethodValue scope methodSubst methodValue
+          Right (foldl X.EApp methodHead evidenceArgs)
+
+    lookupNullaryEvidence deferred methodInfo classArgView =
+      case
+        lookupEvidenceMethodByClass
+          scope
+          (methodInfoOwnerClassSymbolIdentity methodInfo)
+          (typeViewIdentity classArgView)
+          (methodName methodInfo)
+      of
+        Just (runtimeName, _) -> Just runtimeName
+        Nothing ->
+          case deferredMethodEvidence deferred of
+            Just DeferredMethodEvidence {deferredMethodEvidenceRuntimeName = runtimeName} ->
+              Just runtimeName
+            _ -> Nothing
 
     inferNullaryMethodClassArgument methodInfo expectedView
       | deferredMethodFullArityFromInfo methodInfo /= 0 = Nothing
--- a/src/MLF/Frontend/Program/Finalize.hs
+++ b/src/MLF/Frontend/Program/Finalize.hs
@@ -897,10 +897,10 @@ resolveDeferredMethods scope deferredMethods = go
   where
     go env term =
       case deferredPlaceholderHeadWithInsts term of
-        Just (name, _)
+        Just (name, headInsts)
           | Just deferred <- Map.lookup name deferredMethods,
             deferredMethodArgCount deferred == 0 ->
-              resolveDeferredNullaryMethod deferred
+              reapplyHeadInsts headInsts <$> resolveDeferredNullaryMethod deferred
         _ ->
           case term of
             X.EVar {} -> Right term
@@ -1276,6 +1276,10 @@ deferredPlaceholderHeadWithInsts = go []
         X.ETyInst inner (X.InstApp ty) -> go (ty : insts) inner
         X.ETyInst inner _ -> go insts inner
         _ -> Nothing
+
+reapplyHeadInsts :: [ElabType] -> ElabTerm -> ElabTerm
+reapplyHeadInsts insts term =
+  foldl X.ETyInst term (map X.InstApp insts)
 
 dataInfoHeadNames :: ElaborateScope -> DataInfo -> [String]
 dataInfoHeadNames scope info =
--- a/src/MLF/Frontend/Program/Finalize.hs
+++ b/src/MLF/Frontend/Program/Finalize.hs
@@ -35,6 +35,7 @@ import MLF.Frontend.Program.Elaborate
   ( ElaborateScope,
     elaborateScopeDataTypes,
     elaborateScopeRuntimeTypes,
+    classInfoForConstraint,
     inferClassArgument,
     lookupEvidenceMethodByClass,
     lowerType,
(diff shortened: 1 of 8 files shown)


src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -953,6 +959,57 @@ resolveDeferredMethods scope deferredMethods = go
...
+      expectedView <-
+        case deferredMethodExpectedResult deferred of
+          Just view -> Right view
+          Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+      let methodInfo = deferredMethodInfo deferred
+      classArgView <-
+        case inferNullaryMethodClassArgument methodInfo expectedView of
+          Just view -> Right view
+          Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+      (instanceInfo, subst) <- resolveMethodInstanceInfoByTypeView scope methodInfo classArgView
```
soulomoon: This path goes straight to instance lookup, so it skips the hidden method evidence that `extendConstraintEvidence` adds for constrained definitions. A generic nullary method use such as `def neutral : Monoid a => a = mempty; def main : Nat = neutral;` currently fails with `ProgramNoMatchingInstance "Monoid" (STVar "a")`, while the analogous non-nullary `Eq a => \x \y eq x y` path works through local evidence. Please check `lookupEvidenceMethodByClass` for `classArgView` before falling back to instance resolution, and cover that generic nullary case in the matrix.

soulomoon (commented):
I found one correctness gap in the nullary method finalization path.

soulomoon:
Review findings require rework at commit `bc87b09780387932f58c1bdc3e18c208a5447574`.

Findings to address:
- Inline comment added in `src/MLF/Frontend/Program/Finalize.hs`: nullary deferred method finalization skips local hidden class evidence, so generic constrained uses like `def neutral : Monoid a => a = mempty` fail with `ProgramNoMatchingInstance "Monoid" (STVar "a")` instead of using the `Monoid a` evidence. The path should check local evidence before falling back to instance lookup and add matrix coverage.

src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -953,6 +962,75 @@ resolveDeferredMethods scope deferredMethods = go
...
+          Just view -> Right view
+          Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+      let methodInfo = deferredMethodInfo deferred
+      classArgView <-
+        case inferNullaryMethodClassArgument methodInfo expectedView of
+          Just view -> Right view
+          Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+      case lookupNullaryEvidence deferred methodInfo classArgView of
+        Just runtimeName -> Right (X.EVar runtimeName)
+        Nothing -> do
```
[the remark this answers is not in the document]
soulomoon: This still has a local-evidence gap for method-local constraints. The current branch now applies `evidenceArgs`, but those args come from `resolveConstraintEvidenceTerms`, which only goes through instance lookup. A constrained nullary use such as `class Pick a { pick : Eq b => Pair a b; }` with `def picked : (Pick a, Eq b) => Pair a b = pick;` still fails with `ProgramNoMatchingInstance "Eq" (STVar "b")` instead of using the local `Eq b` evidence. Please route these nullary method-local constraints through local evidence before falling back to instances, and cover that polymorphic case in the matrix.

soulomoon:
Review findings require rework at commit `57d428d342bd4ee6ac1b20dd96c36aab98dab81d`.

Findings to address:
- Unresolved review thread: This still has a local-evidence gap for method-local constraints. The current branch now applies `evidenceArgs`, but those args come from `resolveConstraintEvidenceTerms`, which only goes through instance lookup. A constrained nullary use such as `class Pick a { pick : Eq b => Pair a b; }` with `def picked : (Pick a, Eq b) => Pair a b = pick;` still fails with `ProgramNoMatchingInstance "Eq" (STVar "b")` instead of using the local `Eq b` evidence. Please route these nullary method-local constraints through local evidence before falling back to instances, and cover that polymorphic case in the matrix.

soulomoon (commented):
No actionable findings. The reviewed head matches 9d24da32e7a503d01d4432d9f449edba1772834e, the PR is mergeable with the current base, the PR plan and linked issue behavior are covered, `git diff --check origin/master...HEAD` passes, the focused nullary-overloaded matrix slice passes, the authoritative fixture prints `Zero`, and the full gate previously passed on this same head.

Reviewed commit: `9d24da32e7a503d01d4432d9f449edba1772834e`

src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -949,10 +962,123 @@ resolveDeferredMethods scope deferredMethods = go
...
+        case inferNullaryMethodClassArgument methodInfo expectedView of
+          Just view -> Right view
+          Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+      case lookupNullaryEvidence deferred methodInfo classArgView of
+        Just evidence -> do
+          methodSubst <-
+            case inferNullaryMethodSubst methodInfo classArgView Map.empty expectedView of
+              Just subst' -> Right subst'
+              Nothing -> Left (ProgramAmbiguousMethodUse (deferredMethodName deferred))
+          evidenceHead <- instantiateLocalMethodEvidence scope methodSubst evidence
```
soulomoon, line 985: This local-evidence branch still loses the parameterized-result case once the constrained alias is instantiated. A minimal repro is `class DefaultBox a { defaultBox : Box a; }`, `def selected : DefaultBox a => Box a = defaultBox;`, `def main : Box Nat = selected;`: the direct `def main : Box Nat = defaultBox;` path prints `Box Zero`, but the alias path fails in Phase 7 with `TCArgumentMismatch`. Please preserve/specialize the local evidence rewrite for parameterized expected results and add matrix coverage for this local-evidence `Box a` case.

soulomoon (commented):
I found one remaining correctness gap in the nullary local-evidence path.

Issue #99, opened 2026-04-29
Add native LLVM toolchain discovery and executable runner helpers


soulomoon:
## Scope
Extend test support for #89 with portable discovery and invocation of the LLVM/native toolchain needed to turn emitted `.ll` into a linked executable, run it, and capture stdout, stderr, and exit status. Keep this focused on harness infrastructure before wiring broad parity rows.

## Acceptance Criteria
- `LLVMToolSupport` or equivalent helpers can discover required tools, create temporary build products, link an executable, run it, and capture outputs deterministically.
- Missing toolchain pieces produce explicit skipped tests rather than failures unrelated to backend behavior.
- Helpers clean up temporary files and work across the supported local CI/developer environments as far as the existing suite permits.
- Existing `llvm-as` and `llc` smoke behavior remains intact.

## Dependencies / Blockers
- No dependency on other #89 sub-issues.
- May need small coordination with existing backend test harness structure.

## Compatibility With Sibling Sub-Issues
This sub-issue owns tool discovery, linking, process execution, and skip mechanics only. It should not decide `.mlfp` result rendering semantics or broaden parity row assertions.

Issue #87, opened 2026-04-29
Design `.mlfp` Monad, IO primitives, and standard library boundary


soulomoon:
## Goal

Add a tracked design/implementation task for a Haskell-style monadic IO surface in `.mlfp`, grounded in an explicit standard library and primitive runtime boundary.

## Current state

- `.mlfp` programs are pure: the CLI reads a source file, checks/evaluates pure `main`, and renders the resulting value.
- The built-in Prelude currently exposes `Nat`, `Option`, `List`, `Eq`, `eq`, `and`, and `id`.
- There is no source-level `Monad` class, no `IO` type constructor, no `Unit` type, and no IO primitives such as `putStrLn` or `readLine`.
- `emit-backend` emits LLVM IR for the supported first-order subset; it does not currently define an executable IO runtime contract.

## Proposed scope

1. Standard library surface
   - Add `Unit`.
   - Add `Functor` / `Applicative` / `Monad` if desired, or start with the minimal `Monad` class.
   - Expose `pure` and `bind` with higher-kinded class parameter support, e.g. `class Monad (m :: * -> *)`.

2. IO type and primitive boundary
   - Introduce opaque `IO : * -> *` as a built-in/program-layer type constructor rather than a normal source ADT.
   - Add primitive operations such as `__io_pure`, `__io_bind`, `putStrLn : String -> IO Unit`, and optionally `readLine : IO String`.
   - Provide a `Monad IO` instance through the standard library surface.

3. Executable semantics
   - Decide whether `run-program` continues to accept pure `main` values, accepts `main : IO Unit`, or supports both modes.
   - Ensure effects happen only when the checked `main` action is executed by the runner, not during ordinary pure normalization.
   - Define result rendering for `main : IO a` if non-`Unit` results are allowed.

4. Backend semantics
   - Initially either reject IO in `emit-backend` with a structured unsupported diagnostic or lower a small primitive set to runtime/LLVM calls.
   - Keep the backend contract explicit so source typing/inference is not changed by backend-only support.

## Design notes

A source-level `Monad` class is useful for expressing sequencing, but it is not sufficient to implement real IO by itself. `IO a` must be opaque and interpreted by the runner/backend; otherwise it is just another pure value. This keeps raw eMLF/xMLF pure and confines effects to the `.mlfp` program/runtime boundary.

## Acceptance criteria

- A documented standard library contract for `Monad` and the initial IO primitives.
- Checker support for the `IO` type constructor and `Monad IO` surface.
[...]

PR #108, opened 2026-04-29
Implement #100

### LINKED ISSUE #100: Define native entrypoint and result-rendering contract
## Scope
Define how an emitted `.mlfp` `main` value becomes a native process result for #89. Add or expose entrypoint shims and runtime declarations needed by linked executables, and specify stdout and exit status comparison against `ProgramSpec` expectations for pure supported programs.

## Acceptance Criteria
- The backend/runtime contract defines the native entrypoint, result rendering for supported values, stdout/stderr expectations, and exit status behavior.
- Tests or golden fixtures cover representative scalar and ADT results at the native boundary.
- Runtime declarations that cannot remain external stubs are either implemented or explicitly blocked before native-run assertions use them.
- README or backend docs describe the native pipeline contract at a user/developer level.

## Dependencies / Blockers
- Can proceed in parallel with native toolchain helper work.
- Should coordinate with #87 if IO runtime behavior is implemented, but pure ProgramSpec parity must remain independently testable.

This sub-issue owns process-entry and result-rendering semantics. It should not implement the broad parity test matrix wiring or advanced feature coverage.

soulomoon:
Closes #100.

## Implementation Plan

---
issue_number: 100
pr_number: 108
branch: codex/issue-100
---

## Evidence
- Issue #100 asks for the native process contract for emitted `.mlfp` `main` values: entrypoint, result rendering, stdout/stderr, exit status, runtime declarations, tests, and docs.
- PR #108 is open from `codex/issue-100` to `master`; the local branch tracks `origin/codex/issue-100` and currently only has the starter commit.
- Current LLVM output retains source-level functions such as `Main__main`, but does not emit a C ABI `i32 @main()` process entrypoint or native result renderer. Runtime declarations currently include `malloc` and `__mlfp_and`; `__mlfp_and` is still an external declaration in raw backend output.

## Implementation Plan
1. Add a native LLVM emission mode without breaking the raw backend API. Keep `renderBackendProgramLLVM` / `renderCheckedProgramLLVM` as raw library-style LLVM, and add native variants such as `renderBackendProgramNativeLLVM` / `renderCheckedProgramNativeLLVM`. Add a CLI/file helper for native emission, preferably a new command like `emit-native`, rather than switching the broad existing `emit-backend` parity surface in this issue.
2. Extend the small LLVM AST/printer only as needed: support vararg declarations for `printf`, and fix `LLVMSwitch` pretty-printing so switch targets use the switch operand type instead of hard-coded `i64`; this enables `i1` dispatch for booleans and renderer parentheses.
3. Generate the native process wrapper in `MLF.Backend.LLVM.Lower`: reserve/diagnose collisions for `main`, `printf`, native string globals, and renderer helper names; append `define i32 @main()` that calls the checked zero-argument monomorphic source main, prints the rendered value plus newline to stdout, writes nothing to stderr on success, and returns exit code `0`.
4. Implement native result rendering for the supported pure subset: `Int`, `Bool`, and first-order ADT values whose fields recursively render as supported `Int`/`Bool`/ADT values. Use existing backend constructor tags and heap layout; derive display constructor names by stripping the runtime module prefix. Generate monomorphic ADT renderer helpers by concrete result type, including ProgramSpec-compatible nested ADT parentheses such as `Some (Succ Zero)`. Reject function, forall, string, unknown data, type-variable, polymorphic, and IO-like results with explicit `BackendLLVMUnsupportedExpression` diagnostics.
5. In native mode, do not leave `__mlfp_and` as an unresolved external when it is used and not user-owned. Emit a small internal LLVM definition for it. Keep `malloc` and `printf` as libc-provided declarations.
6. Add focused `BackendLLVMSpec` coverage for native `Int`, `Bool`, nullary/unary/nested ADT output, and a Prelude `and` case proving the runtime primitive is linkable/executable. Compare expected stdout as the ProgramSpec value plus `\n`, stderr as empty, and exit status as success. Use existing `llvm-as`/`llc` validation and add only a narrow `lli` execution helper with pending behavior if `lli` is absent; leave full compile/link runner infrastructure to #99/#101.
7. Add negative tests for unsupported native result types and native name collisions so native-run assertions cannot silently target unsupported rows.
8. Update README and backend architecture docs with the native contract: raw backend vs native emission, source main retention, generated `i32 @main`, stdout newline/value rendering, no stderr on success, exit `0`, supported/unsupported result types, libc/runtime declarations, and IO remaining out of scope for #87.
9. Validate with a focused backend LLVM test slice first, then run `cabal build all && cabal test` before publishing. Before committing, inspect `git status --short` and relevant diffs; stage only issue-related code/docs/tests, commit on `codex/issue-100`, and push to PR #108.

---

### DIFF
--- a/src/MLF/Backend/LLVM/Lower.hs
+++ b/src/MLF/Backend/LLVM/Lower.hs
@@ -18,7 +18,7 @@ import Control.Monad (foldM, unless, when, zipWithM, zipWithM_)
 import Control.Monad.State.Strict (StateT (StateT), evalStateT, get, gets, modify)
 import Data.Bifunctor (first)
 import Data.Char (isAlphaNum, ord)
-import Data.List (intercalate, isPrefixOf, nub, sort, sortOn)
+import Data.List (intercalate, isPrefixOf, nub, sort, sortOn, stripPrefix)
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NE
 import Data.Map.Strict (Map)
@@ -317,9 +317,12 @@ constructorNameGlobals base spec =
   case nativeDataRuntimeForTypeName (nrsType spec) of
     Nothing -> []
     Just dataName ->
-      [ LLVMStringGlobal (nativeConstructorGlobalName spec constructorRuntime) (displayConstructorName (backendConstructorName (crConstructor constructorRuntime)))
-      | constructorRuntime <- maybe [] drConstructors (Map.lookup dataName (pbData base))
-      ]
+      case Map.lookup dataName (pbData base) of
+        Nothing -> []
+        Just dataRuntime0 ->
+          [ LLVMStringGlobal (nativeConstructorGlobalName spec constructorRuntime) (displayConstructorName dataRuntime0 constructorRuntime)
+          | constructorRuntime <- drConstructors dataRuntime0
+          ]
 
 nativeDataRuntimeForTypeName :: BackendType -> Maybe String
 nativeDataRuntimeForTypeName =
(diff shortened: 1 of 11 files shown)


src/MLF/Backend/LLVM/Lower.hs
```diff
@@ -172,14 +208,442 @@ lowerBackendProgram program = do
...
+    "" -> runtimeName
+    name -> name
+
+stripModulePrefix :: String -> String
+stripModulePrefix name =
+  go name
+  where
+    go remaining =
+      case breakDoubleUnderscore remaining of
+        Just rest -> go rest
```
soulomoon: This recurses over every `__` in the runtime constructor name, so it strips user-authored double underscores inside the constructor itself. Constructor identifiers can contain underscores; for example `data Weird = A__B : Weird; def main : Weird = A__B;` prints `A__B` via `run-program`, but `emit-native`/`lli` prints `B`. That breaks the PR's ProgramSpec-compatible ADT rendering contract. Strip only the module/runtime prefix, or carry the source display name through the constructor metadata.

soulomoon (commented):
Found one native rendering correctness issue.

soulomoon:
Review findings require rework at commit `158120a8e59577b2c407657d4f1c4de53dbccc61`.

Findings to address:
- Native ADT result rendering strips every double-underscore segment from constructor runtime names, so a supported constructor like `A__B` renders as `B` instead of ProgramSpec's `A__B`; I added an inline comment on `src/MLF/Backend/LLVM/Lower.hs:636`.

soulomoon (commented):
LGTM. The PR remains at the target commit, is mergeable, satisfies the native entrypoint/result-rendering plan and issue #100 acceptance criteria, and the native-focused test slice passes.

Reviewed commit: `a2b868f8a867e978ec2a0a0d4930a2b7e6dc5a97`

PR #109, opened 2026-04-30
Implement #9

### LINKED ISSUE #9: Support expected-type resolution for nullary overloaded .mlfp methods

soulomoon:
Closes #9.

## Implementation Plan

---
issue_number: 9
pr_number: 109
branch: codex/issue-9-2
---

## Context
- Issue #9 is open and PR #109 is open against `master` from `codex/issue-9-2`.
- Local worktree is clean on `codex/issue-9-2`; the branch currently only contains the watcher start commit on top of `origin/HEAD`.
- The rework target is the post-merge finding: `def selected : DefaultBox a => Box a = defaultBox; def main : Box Nat = selected` fails in Phase 7 with `TCArgumentMismatch`, while direct `def main : Box Nat = defaultBox` already works.

## Implementation Plan
1. Add a regression row in `test/Parity/ProgramMatrix.hs` near the existing nullary overloaded method cases for the parameterized local-evidence alias path: `DefaultBox a`, `Box a`, `selected : DefaultBox a => Box a = defaultBox`, and `main : Box Nat = selected`, expecting runtime value `Box Zero`.
2. Fix `src/MLF/Frontend/Program/Finalize.hs` in the nullary deferred-method rewrite so placeholder head type instantiations are preserved only where they are semantically needed. Specifically, pass the collected head instantiations into nullary resolution and keep the existing preservation for global instance-method replacement, but avoid replaying placeholder instantiations over already-typed local evidence parameters. Local evidence should be instantiated from the recovered method substitution and its own evidence type via `instantiateLocalMethodEvidence`, not by blindly reapplying the placeholder spine.
3. Keep ambiguity fail-closed: no expected result should still produce `ProgramAmbiguousMethodUse`, and missing/no matching evidence should continue through the existing `ProgramNoMatchingInstance`/ambiguous paths.
4. Update `implementation_notes.md` and `CHANGELOG.md` with a short 2026-04-30 note describing the constrained-alias fix for parameterized nullary associated values. Do not change `docs/mlfp-language-reference.md` unless implementation changes the already-documented contract.
5. Validate with a focused Hspec match for the new case, then run nearby nullary overloaded method matrix coverage. Before claiming completion, run `cabal build all && cabal test` as the required behavior-changing full gate.
6. Before publishing, inspect `git status --short` and relevant diffs, stage only issue-related files, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, run `gh auth status` and `gh auth setup-git`, then push to the existing `codex/issue-9-2` branch for PR #109. Do not create another PR or resolve review threads.

---

### DIFF
--- a/test/Parity/ProgramMatrix.hs
+++ b/test/Parity/ProgramMatrix.hs
@@ -407,6 +407,31 @@ emlfBoundaryMatrix =
         )
         (ExpectRunValue "Box Zero")
     , ProgramMatrixCase
+        "checks parameterized nullary overloaded method through local evidence alias"
+        ( InlineProgram $
+            unlines
+                [ "module Main export (DefaultBox, Nat(..), Box(..), defaultBox, selected, main) {"
+                , "  class DefaultBox a {"
+                , "    defaultBox : Box a;"
+                , "  }"
+                , ""
+                , "  data Nat ="
+                , "      Zero : Nat;"
+                , ""
+                , "  data Box a ="
+                , "      Box : a -> Box a;"
+                , ""
+                , "  instance DefaultBox Nat {"
+                , "    defaultBox = Box Zero;"
+                , "  }"
+                , ""
+                , "  def selected : DefaultBox a => Box a = defaultBox;"
+                , "  def main : Box Nat = selected;"
+                , "}"
+                ]
+        )
+        ExpectCheckSuccess
+    , ProgramMatrixCase
         "rejects nullary overloaded method without expected type evidence"
         ( InlineProgram $
             unlines
--- a/test/Parity/ProgramMatrix.hs
+++ b/test/Parity/ProgramMatrix.hs
@@ -407,7 +407,7 @@ emlfBoundaryMatrix =
         )
         (ExpectRunValue "Box Zero")
     , ProgramMatrixCase
-        "checks parameterized nullary overloaded method through local evidence alias"
+        "runs parameterized nullary overloaded method through local evidence alias"
         ( InlineProgram $
             unlines
                 [ "module Main export (DefaultBox, Nat(..), Box(..), defaultBox, selected, main) {"
@@ -430,7 +430,7 @@ emlfBoundaryMatrix =
                 , "}"
                 ]
         )
-        ExpectCheckSuccess
+        (ExpectRunValue "Box Zero")
     , ProgramMatrixCase
         "rejects nullary overloaded method without expected type evidence"
         ( InlineProgram $
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@
   propagated expected source type while bare uses without such evidence still
   fail closed as ambiguous. Parameterized associated values used through
   constrained aliases, such as `DefaultBox a => Box a`, now preserve local
-  evidence and expected result annotations without over-instantiating the
-  rewrite.
+  evidence, expected result annotations, and recovered ADT instantiation
+  identity without over-instantiating the rewrite.
 - Added a native LLVM emission contract for pure `.mlfp` entrypoints. The
   backend can now emit an `i32 @main()` wrapper that renders supported
   `Int`, `Bool`, and first-order ADT results to stdout, returns exit status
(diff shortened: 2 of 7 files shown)


test/Parity/ProgramMatrix.hs
```diff
@@ -406,6 +406,31 @@ emlfBoundaryMatrix =
...
+                , "  instance DefaultBox Nat {"
+                , "    defaultBox = Box Zero;"
+                , "  }"
+                , ""
+                , "  def selected : DefaultBox a => Box a = defaultBox;"
+                , "  def main : Box Nat = selected;"
+                , "}"
+                ]
+        )
+        ExpectCheckSuccess
```
soulomoon: The PR plan calls for this new matrix row to expect the runtime value `Box Zero`, but `ExpectCheckSuccess` only verifies type checking. The runtime matrix rows are also what feed the shared interpreter/LLVM parity coverage, so this leaves the local-evidence alias path out of that coverage. Please make this row `ExpectRunValue "Box Zero"` (and drop the standalone duplicate if it becomes redundant).

soulomoon (commented):
Found one test-coverage/plan mismatch.

soulomoon:
Review findings require rework at commit `5e12abd32b3123f5d237668afffb2e32a8554159`.

Findings to address:
- test/Parity/ProgramMatrix.hs:433 adds the requested local-evidence alias matrix case as ExpectCheckSuccess instead of the PR-plan-required ExpectRunValue "Box Zero", leaving that path out of the shared runtime/LLVM parity matrix coverage.

soulomoon (commented):
Clean review. The current head matches the target commit, the issue-required local-evidence alias path has runtime parity coverage with `ExpectRunValue "Box Zero"`, and the implementation satisfies the PR plan with no new actionable findings.

Reviewed commit: `05057916afe2f5acab3bc336222c1f468559c6b7`

PR #111, opened 2026-04-30
Implement #91

### LINKED ISSUE #91: Add source typing for Unit, IO, IO primitives, and Monad IO
## Scope
Implement the checked source surface chosen by the #87 contract: `Unit`, opaque `IO : * -> *`, the initial primitive operations, and the `Monad IO` standard-library instance or equivalent built-in binding surface. Keep `IO` opaque at the program boundary rather than encoding it as an ordinary user ADT.

## Acceptance Criteria
- Positive `.mlfp` tests check programs that mention `Unit`, construct or reference IO actions through the selected primitives, and typecheck `pure`/`bind` for `IO`.
- Negative tests reject attempts to inspect or pattern match the opaque `IO` representation.
- Method and instance resolution for `Monad IO` is coherent and does not weaken existing typeclass ambiguity checks.
- Existing pure Prelude bindings continue to typecheck and run unchanged.

## Dependencies / Blockers
- Depends on `Specify the .mlfp IO and Monad standard library contract`.
- Blocks runner execution and backend behavior sub-issues for #87.

This sub-issue owns source typing and standard-library exposure only. It should not execute effects in `run-program` and should not add LLVM lowering beyond enough metadata for later siblings.

soulomoon:
Closes #91.

## Implementation Plan

---
issue_number: 91
pr_number: 111
branch: codex/issue-91
---

## Implementation Plan

1. Reconfirm the implementation turn starts on `codex/issue-91` for PR #111 with a clean status, then read the watcher-written issue plan before editing.

2. Add a small centralized built-in program-surface helper for built-in symbols, type kinds, and primitive value types. Keep existing `Int`, `Bool`, and `String` as kind `*`; add opaque `IO` with kind `* -> *`; add reserved primitive values `__io_pure`, `__io_bind`, and `__io_putStrLn` with source types `forall a. a -> IO a`, `forall a b. IO a -> (a -> IO b) -> IO b`, and `String -> IO Unit`.

3. Represent `IO` as an explicit opaque built-in type, not Church-encoded data. Synthesize it for the built-in `Prelude` export/import path, give it a stable identity such as `<builtin>.IO`, reject user attempts to define an ordinary `IO` data type, allow abstract `IO` imports/exports, and keep `IO(..)`/constructor imports failing because there are no constructors.

4. Update the built-in Prelude source to export `Unit(..)`, `IO`, `Monad`, `pure`, `bind`, and `putStrLn`; add ordinary `Unit`; add `class Monad (m :: * -> *)` with `pure` and `bind`; add `instance Monad IO` backed by `__io_pure`/`__io_bind`; add public `putStrLn` backed by `__io_putStrLn`. Preserve existing Prelude exports and behavior for `Nat`, `Option`, `List`, `Eq`, `eq`, `and`, and `id`.

5. Extend overloaded method resolution narrowly so non-nullary methods can use an expected result type when arguments alone do not determine the class parameter. This is needed for `def x : IO Unit = pure Unit;`. Store the expected result view in `DeferredMethodCall`, combine argument-derived substitutions with result-template matching, and keep existing ambiguity behavior when neither arguments nor expected type determine the instance.

6. Keep opaque `IO` out of representation-sensitive paths: do not generate constructor bindings for it, do not lower it as a source ADT, do not allow case/constructor inspection of it, and avoid backend/runtime semantic changes beyond any metadata filtering needed so existing pure programs still run and lower unchanged.

7. Add focused tests in `ProgramSpec` or the existing program matrix: positive checks for importing `Unit`, `IO`, `Monad`, `pure`, `bind`, and `putStrLn`; `IO Unit` annotations; `pure Unit`; `bind (putStrLn "...") (\(_done : Unit) putStrLn "...")`; and coherent `Monad IO` instance selection. Add negative tests for `IO(..)` import/constructor inspection and an ambiguity guard showing `pure` is still rejected when no expected `IO` result fixes the instance.

8. Update relevant docs: README Prelude surface list and `.mlfp` language reference only where the implemented behavior differs from the already documented #87 contract; add a concise CHANGELOG entry because this is meaningful source-language progress.

9. Validate with a focused ProgramSpec slice first, then run `cabal build all && cabal test`. Before committing, inspect `git status --short` and related diffs, stage only issue-related files, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, run `gh auth status`/`gh auth setup-git` before publishing, and push `codex/issue-91` to the existing PR #111.

---

### DIFF
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -45,6 +45,7 @@ import MLF.Elab.Types
     schemeFromType,
     tyToElab,
   )
+import qualified MLF.Frontend.Program.Builtins as Builtins
 import MLF.Frontend.Program.Elaborate (ElaborateScope, elaborateScopeDataTypes, lowerType, mkElaborateScope)
 import MLF.Frontend.Program.Types
   ( CheckedBinding (..),
@@ -116,6 +117,7 @@ type LiftM = StateT LiftState (Either BackendConversionError)
 
 convertCheckedProgram :: CheckedProgram -> Either BackendConversionError BackendProgram
 convertCheckedProgram checked = do
+  rejectOpaqueBuiltinMain checked
   context <- buildConvertContext checked
   initialEnv <- buildInitialEnv context checked
   modules0 <- mapM (convertCheckedModule context initialEnv) (checkedProgramModules checked)
@@ -184,13 +186,35 @@ backendBuiltinTermTypes =
 convertCheckedModule :: ConvertContext -> Env -> CheckedModule -> Either BackendConversionError BackendModule
 convertCheckedModule context env checkedModule = do
   dataDecls <- mapM (convertDataInfo context) (Map.elems (checkedModuleData checkedModule))
-  bindings <- concat <$> mapM (convertCheckedBinding context env checkedModule) (checkedModuleBindings checkedModule)
+  bindings <-
+    concat
+      <$> mapM
+        (convertCheckedBinding context env checkedModule)
+        (filter (not . checkedBindingMentionsOpaqueBuiltin) (checkedModuleBindings checkedModule))
   Right
     BackendModule
       { backendModuleName = checkedModuleName checkedModule,
         backendModuleData = dataDecls,
         backendModuleBindings = bindings
       }
+
+rejectOpaqueBuiltinMain :: CheckedProgram -> Either BackendConversionError ()
+rejectOpaqueBuiltinMain checked =
+  case
+    [ binding
+      | checkedModule <- checkedProgramModules checked,
+        binding <- checkedModuleBindings checkedModule,
+        checkedBindingName binding == checkedProgramMain checked
+    ]
+  of
+    binding : _
+      | checkedBindingMentionsOpaqueBuiltin binding ->
+          Left (BackendUnsupportedCaseShape "IO programs are not supported by backend conversion yet")
+    _ -> Right ()
+
+checkedBindingMentionsOpaqueBuiltin :: CheckedBinding -> Bool
+checkedBindingMentionsOpaqueBuiltin =
+  Builtins.srcTypeMentionsOpaqueBuiltin . checkedBindingSourceType
 
 convertCheckedBinding :: ConvertContext -> Env -> CheckedModule -> CheckedBinding -> Either BackendConversionError [BackendBinding]
 convertCheckedBinding context env checkedModule binding = do
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -210,7 +210,38 @@ rejectOpaqueBuiltinMain checked =
     binding : _
       | checkedBindingMentionsOpaqueBuiltin binding ->
           Left (BackendUnsupportedCaseShape "IO programs are not supported by backend conversion yet")
-    _ -> Right ()
+    _ ->
+      case referencedOpaqueBuiltinBindings checked of
+        [] -> Right ()
+        refs ->
+          Left
+            ( BackendUnsupportedCaseShape
+                ( "IO dependencies are not supported by backend conversion yet: "
+                    ++ intercalate ", " [owner ++ " -> " ++ opaque | (owner, opaque) <- refs]
+                )
+            )
+
+referencedOpaqueBuiltinBindings :: CheckedProgram -> [(String, String)]
+referencedOpaqueBuiltinBindings checked =
+  [ (checkedBindingName binding, opaqueName)
+    | binding <- allCheckedBindings checked,
+      not (checkedBindingMentionsOpaqueBuiltin binding),
+      opaqueName <- Set.toList (freeTermVariables (checkedBindingTerm binding) `Set.intersection` opaqueNames)
+  ]
+  where
+    opaqueNames =
+      Set.fromList
+        [ checkedBindingName binding
+          | binding <- allCheckedBindings checked,
+            checkedBindingMentionsOpaqueBuiltin binding
+        ]
+
+allCheckedBindings :: CheckedProgram -> [CheckedBinding]
+allCheckedBindings checked =
+  [ binding
+    | checkedModule <- checkedProgramModules checked,
+      binding <- checkedModuleBindings checkedModule
+  ]
 
 checkedBindingMentionsOpaqueBuiltin :: CheckedBinding -> Bool
 checkedBindingMentionsOpaqueBuiltin =
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -230,7 +230,8 @@ referencedOpaqueBuiltinBindings checked =
   ]
   where
     opaqueNames =
-      Set.fromList
+      Builtins.builtinOpaqueValueNames
+        `Set.union` Set.fromList
         [ checkedBindingName binding
           | binding <- allCheckedBindings checked,
             checkedBindingMentionsOpaqueBuiltin binding
--- a/src/MLF/Frontend/Program/Elaborate.hs
+++ b/src/MLF/Frontend/Program/Elaborate.hs
@@ -1428,7 +1428,9 @@ annotateExpectedValueUse scope mbExpected valueInfo applied =
   case mbExpected of
     Just expectedTy
       | not (isLocalOrdinaryValue valueInfo),
-        isRecursiveResultType expectedTy || isRecursiveResultType (lowerType scope expectedTy) ->
+        isRecursiveResultType expectedTy
+          || isRecursiveResultType (lowerType scope expectedTy)
+          || Builtins.srcTypeMentionsOpaqueBuiltin expectedTy ->
           surfaceAnn applied (lowerType scope expectedTy)
     _ -> applied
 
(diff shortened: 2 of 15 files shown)


src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -83,6 +86,21 @@ import MLF.Frontend.Program.Types
 import MLF.Frontend.Syntax (Expr (..), SrcBound (..), SrcTy (..), SrcType, SurfaceExpr)
 import MLF.Reify.TypeOps (alphaEqType, churchAwareEqType, freeTypeVarsType)
 
+finalizeBindingAllowOpaque :: ElaborateScope -> LoweredBinding -> Either ProgramError CheckedBinding
+finalizeBindingAllowOpaque scope lowered
+  | Builtins.srcTypeMentionsOpaqueBuiltin (loweredBindingSourceType lowered) = do
```
soulomoon, line 91: This branch is skipping the normal surface pipeline and deferred-obligation finalization for every binding whose source type mentions `IO`, so the RHS is never checked against the annotation. I verified that `def main : IO Unit = 1;` currently returns `Right` from `checkLocatedProgram (withPreludeLocated ...)`, even though an `Int` literal is not an `IO Unit` action. Please still validate the lowered RHS against the expected type (and resolve deferred methods) before replacing the checked term with an opaque placeholder.

soulomoon (commented):
Found one correctness issue in the opaque IO finalization path.

soulomoon:
Review findings require rework at commit `18aad0cd05b72c2cdfd1be7246708d4096b164d8`.

Findings to address:
- `finalizeBindingAllowOpaque` bypasses normal RHS typechecking and deferred-obligation finalization for any binding whose source type mentions `IO`; I verified `def main : IO Unit = 1;` is accepted by `checkLocatedProgram`, so non-IO expressions can be checked as IO actions.

src/MLF/Frontend/Program/Elaborate.hs
```diff
@@ -2068,48 +2100,72 @@ registerDeferredObligation placeholder placeholderTy obligation =
...
       methodTy = methodType methodInfo
       quantifiedMethodTy = quantifiedMethodType methodInfo
-      knownClassArg = knownResolvedMethodClassArg scope methodInfo args
+      knownClassArg = knownResolvedMethodClassArg scope methodInfo args mbExpectedResult
    in case knownClassArg of
-        Just classArgTy -> stripVacuousSrcForalls (specializeMethodType methodTy paramName classArgTy)
+        Just classArgTy ->
+          let specializedTy = stripVacuousSrcForalls (specializeMethodType methodTy paramName classArgTy)
+              callSubst =
+                case inferResolvedMethodCallSubst scope methodInfo classArgTy args of
```
soulomoon, line 2129: This `callSubst` path still misses type variables that are determined by polymorphic method arguments, so the expected result can select `Monad IO` while different arguments instantiate the method variable inconsistently. For example, `def main : IO Unit = bind (__io_pure 1) (\(_n : Unit) putStrLn "world");` currently typechecks even though the first argument fixes `a ~ Int` and the continuation requires `a ~ Unit`. Please derive/check the full method substitution from all supplied arguments before accepting the expected-result match.

soulomoon (commented):
Found one remaining correctness issue in the expected-result method resolution path.

soulomoon:
Review findings require rework at commit `798b3f061afc4839627b48ae43a8ec4963ee7fc4`.

Findings to address:
- Expected-result resolution for `bind` can select `Monad IO` without enforcing a consistent substitution across supplied arguments; I verified `def main : IO Unit = bind (__io_pure 1) (\(_n : Unit) putStrLn "world");` currently typechecks even though the first argument fixes `a ~ Int` and the continuation requires `a ~ Unit`.

src/MLF/Frontend/Program/Run.hs
```diff
@@ -36,30 +40,110 @@ data Value
...
+      checkedBindingName binding `Set.member` reachableNames,
+      not (checkedBindingMentionsOpaqueBuiltin binding)
+  ]
+  where
+    reachableNames = reachableBindingNames checked (Set.singleton (checkedProgramMain checked))
+
+reachableOpaqueRuntimeBindings :: CheckedProgram -> [CheckedBinding]
+reachableOpaqueRuntimeBindings checked =
+  [ binding
+    | binding <- allCheckedBindings checked,
```
soulomoon, line 100: This only treats reachable top-level bindings whose source type mentions `IO` as unsupported, so direct uses of the new primitive globals still slip through. For example, `def main : Unit = (\(_action : IO Unit) Unit) (__io_pure Unit);` checks and `runLocatedProgram` returns a residual term containing free `__io_pure` instead of failing closed. Please include the opaque primitive runtime names in the dependency check, or hide them from user modules outside the built-in Prelude path.

src/MLF/Backend/Convert.hs
```diff
@@ -184,14 +186,67 @@ backendBuiltinTermTypes =
...
+                    ++ intercalate ", " [owner ++ " -> " ++ opaque | (owner, opaque) <- refs]
+                )
+            )
+
+referencedOpaqueBuiltinBindings :: CheckedProgram -> [(String, String)]
+referencedOpaqueBuiltinBindings checked =
+  [ (checkedBindingName binding, opaqueName)
+    | binding <- allCheckedBindings checked,
+      not (checkedBindingMentionsOpaqueBuiltin binding),
+      opaqueName <- Set.toList (freeTermVariables (checkedBindingTerm binding) `Set.intersection` opaqueNames)
```
soulomoon, line 229: Same gap here: `opaqueNames` only contains checked bindings, not the reserved primitive globals. A pure `main` that directly mentions `__io_pure` reaches backend conversion and fails later as `TCUnboundVar "__io_pure"`, instead of the intended IO fail-closed diagnostic. The primitive IO names should be rejected by this guard as well, or scoped so user modules cannot reference them directly.

soulomoon (commented):
Found a remaining fail-closed gap for direct IO primitive references.

soulomoon:
Review findings require rework at commit `0b91e7742ad96c89008f3b3b18539e2a10be1dc9`.

Findings to address:
- Direct uses of the reserved IO primitive globals are not included in `run-program`'s opaque dependency guard; I verified `def main : Unit = (\(_action : IO Unit) Unit) (__io_pure Unit);` checks and `runLocatedProgram` returns a residual term containing free `__io_pure` instead of failing closed.
- Backend conversion has the same direct-primitive gap; the same pure `main` reaches conversion and fails later with `TCUnboundVar "__io_pure"` rather than the intended IO fail-closed diagnostic.

soulomoon (commented):
Found one correctness issue in the opaque IO fallback.

src/MLF/Frontend/Program/Finalize.hs
```diff
@@ -80,9 +83,96 @@ import MLF.Frontend.Program.Types
...
+        Nothing
+          | opaqueSourceCompatible scope argTy paramTy -> Right resultTy
+          | otherwise -> Left (ProgramTypeMismatch argTy paramTy)
+    other -> Left (ProgramExpectedFunction other)
+
+opaqueSourceCompatible :: ElaborateScope -> SrcType -> SrcType -> Bool
+opaqueSourceCompatible scope actualTy expectedTy =
+  alphaEqSrcType actualTy expectedTy
+    || alphaEqSrcType (lowerType scope actualTy) (lowerType scope expectedTy)
+    || sourceForallMatches expectedTy actualTy
```
soulomoon: Using `sourceForallMatches` here makes the opaque fallback accept monomorphic actions for polymorphic IO annotations. I verified `def main : forall a. IO a = __io_pure 1;` returns `Right` from `checkLocatedProgram (withPreludeLocated ...)`, because the fallback infers `IO Int` and then drops the expected `forall a` while matching `IO a`. Please keep this fallback from instantiating expected `forall`s when validating an annotated binding; `forall a. IO a` should require a genuinely polymorphic RHS, not a single `IO Int` action.

soulomoon:
Review findings require rework at commit `ca2abb1bfe4455c280c8c3cffae16920c3630f79`.

Findings to address:
- `src/MLF/Frontend/Program/Finalize.hs:165` allows the opaque fallback to accept monomorphic IO actions for polymorphic IO annotations; verified `def main : forall a. IO a = __io_pure 1;` currently typechecks.
- `docs/mlfp-language-reference.md` still has a stale "Current prelude contents" list that omits the newly implemented `Unit`, `IO`, `Monad`, `pure`, `bind`, and `putStrLn` surface from the PR plan.

soulomoon:
Review findings require rework at commit `cc1394f1df215a2b3904cd49227976924741bca1`.

Findings to address:
- The current head still does not satisfy the issue/PR-plan primitive surface for `__io_bind`. I verified `__io_pure` and `__io_putStrLn` can be used directly, but a well-typed direct bind primitive program (`def main : IO Unit = __io_bind (__io_pure Unit) (\(_n : Unit) __io_putStrLn "world");`) is rejected by `checkLocatedProgram (withPreludeLocated ...)` with `PhiInvariantError "PhiReorder: missing binder identity at positions [0,1]"`. Please either make the direct `__io_bind` primitive typecheck consistently with the other IO primitives, or hide the reserved primitives from user modules and adjust the acceptance/docs surface accordingly.

soulomoon (commented):
Clean review. The current head matches the PR plan and linked issue acceptance criteria, and the validation already run for this same target commit included focused IO checks plus `cabal build all && cabal test` passing.

Reviewed commit: `4d326d095a6f44457ecbf836f32078c85221cd02`

PR #116, opened 2026-04-30
Implement #96

### LINKED ISSUE #96: Compile partial applications as closure values
## Scope
Support underapplied global, local, and known function calls in the LLVM backend by constructing closure values that capture supplied arguments and apply remaining arguments later. Preserve existing saturated-call behavior and source-level arity diagnostics.

## Acceptance Criteria
- Backend rejection specs for partial applications become LLVM success specs when source checking accepts the program.
- Tests cover partial application of top-level functions, local functions, and methods or helpers after typeclass resolution when applicable.
- Generated LLVM validates with `llvm-as` and representative partial-application cases pass object-code smoke checks.
- Overapplication and genuinely ambiguous overloaded partial uses still fail through existing frontend/checker rules.

## Dependencies / Blockers
- Depends on `Define the LLVM closure ABI and backend IR representation` and `Closure-convert escaping lambdas and indirect calls`.

This sub-issue owns partial application packaging. It should reuse the closure ABI and should not change ADT field representation or recursive closure promotion except where needed for direct partial-application fixtures.

soulomoon:
Closes #96.

## Implementation Plan

---
issue_number: 96
pr_number: 116
branch: codex/issue-96
---

## Implementation Plan

1. Reconfirm branch/worktree before editing: stay on `codex/issue-96`, confirm PR #116 is the existing PR, and keep watcher-owned files untouched.

2. Add focused failing coverage first in `test/BackendLLVMSpec.hs` and/or `test/Parity/ProgramMatrix.hs` for checked-source partial applications that currently should lower through LLVM:
   - top-level known function partial application, then later closure call;
   - local let-bound function partial application, then later closure call;
   - keep the existing overloaded/typeclass partial rows green, and add representative object-code smoke coverage for at least one partial-application row.
   Convert the current raw partial-application rejection expectation into a success fixture using the explicit closure shape or checked-source path; avoid inventing a second raw underapplied-`BackendApp` ABI.

3. Implement partial-application packaging in `src/MLF/Backend/Convert.hs` using the existing closure IR:
   - recognize `collectApps` shapes where the checked head type has more value parameters than supplied arguments and the remaining type matches the expected result function type;
   - leave saturated calls on the existing direct/closure-call paths and leave overapplication/frontend ambiguity to existing checker diagnostics;
   - generate a `BackendClosure` with captures for supplied arguments and, when needed, the closure-valued callee, plus fresh remaining value parameters;
   - make the closure body apply captured supplied arguments followed by remaining parameters, using `BackendClosureCall` for closure-valued heads and direct backend applications for known direct global calls.

4. Make local-function underapplication work without weakening saturated local calls:
   - extend the local let closure-demand analysis so a let-bound function is closure-converted when it is used bare, passed/returned as a value, or underapplied;
   - keep fully saturated local calls as direct local calls.

5. Touch `src/MLF/Backend/LLVM/Lower.hs` only if tests expose a lowering gap for direct `BackendClosure` values used as first-order function arguments or for closure-entry collection/qualification through the new shapes. Reuse the existing closure ABI and validation instead of adding a new runtime representation.

6. Update durable docs for the behavior change: `docs/architecture.md`, `implementation_notes.md`, and `CHANGELOG.md` if the completed change records meaningful project progress.

7. Validate in increasing scope:
   - targeted partial/LLVM specs, for example `cabal test mlf2-test --test-show-details=direct --test-options='--match "partial"'` and an `MLF.Backend.LLVM` slice;
   - ensure generated LLVM validates with the existing `llvm-as` helpers and object-code smoke cases where tools are available;
   - final gate before completion: `cabal build all && cabal test`.

8. Publish only after validation: run `gh auth status` and `gh auth setup-git`, inspect `git status --short` and relevant diffs, stage only issue-related files, commit as `codex-watcher <codex-watcher@users.noreply.github.com>`, and push `codex/issue-96` to the existing PR #116 without creating another PR.

---

### DIFF
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -1410,10 +1410,11 @@ aliasedClosureValueArguments context scope bindingTy term =
   case stripClosureHeadTypeInsts headTerm of
     EVar name ->
       Set.fromList
-        [ demandedIndex - suppliedCount
+        [ paramOffset + exposedIndex
         | demandedIndex <- Set.toList (lookupClosureValueArgumentDemand context scope name),
+          let exposedIndex = demandedIndex - suppliedCount,
           demandedIndex >= suppliedCount,
-          demandedIndex - suppliedCount < exposedCount
+          exposedIndex < exposedCount
         ]
     _ ->
       Set.empty
@@ -1421,6 +1422,7 @@ aliasedClosureValueArguments context scope bindingTy term =
     (_, valueTy) = splitBackendForalls bindingTy
     (paramTys, _) = splitBackendArrows valueTy
     (params, body) = collectLeadingLams (length paramTys) term
+    paramOffset = length params
     exposedCount = length paramTys - length params
     (headTerm, suppliedArgs) = collectAliasedApps body
     suppliedCount = length suppliedArgs
--- a/src/MLF/Backend/Convert.hs
+++ b/src/MLF/Backend/Convert.hs
@@ -134,6 +134,10 @@ data LambdaMode
   = DirectLambda
   | ClosureLambda (Maybe String)
 
+data PartialApplicationMode
+  = AllowPartialApplications
+  | SuppressPartialApplications
+
 emptyClosureScope :: ClosureScope
 emptyClosureScope =
   ClosureScope
@@ -1781,22 +1785,31 @@ convertTerm context env scope =
   convertTermExpectedMode DirectLambda context env scope Nothing
 
 convertTermExpectedMode :: LambdaMode -> ConvertContext -> Env -> ClosureScope -> Maybe BackendType -> ElabTerm -> ConvertM BackendExpr
-convertTermExpectedMode mode context env scope mbExpectedTy term =
+convertTermExpectedMode =
+  convertTermExpectedModeWith AllowPartialApplications
+
+convertTermExpectedModeNoPartial :: LambdaMode -> ConvertContext -> Env -> ClosureScope -> Maybe BackendType -> ElabTerm -> ConvertM BackendExpr
+convertTermExpectedModeNoPartial =
+  convertTermExpectedModeWith SuppressPartialApplications
+
+convertTermExpectedModeWith :: PartialApplicationMode -> LambdaMode -> ConvertContext -> Env -> ClosureScope -> Maybe BackendType -> ElabTerm -> ConvertM BackendExpr
+convertTermExpectedModeWith partialMode mode context env scope mbExpectedTy term =
   case mbExpectedTy of
     Just resultTy0 ->
       let resultTy = canonicalizeBackendType context resultTy0
-       in convertSpecialTerm mode context env scope term resultTy
+       in convertSpecialTerm partialMode mode context env scope term resultTy
             >>= \case
               Just expr -> pure expr
               Nothing -> convertOrdinaryTerm mode context env scope term resultTy
     Nothing -> do
       resultTy <- canonicalizeBackendType context <$> liftEitherConvert (inferBackendType env term)
-      convertSpecialTerm mode context env scope term resultTy
+      convertSpecialTerm partialMode mode context env scope term resultTy
         >>= \case
           Just expr -> pure expr
           Nothing -> convertOrdinaryTerm mode context env scope term resultTy
 
 convertSpecialTerm ::
+  PartialApplicationMode ->
   LambdaMode ->
   ConvertContext ->
   Env ->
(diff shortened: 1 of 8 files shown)


src/MLF/Backend/Convert.hs
```diff
@@ -1781,35 +1785,217 @@ convertTerm context env scope =
...
+          headTy <- normalizeBackendTypeForContext context <$> liftEitherConvert (inferBackendType env headTerm)
+          let (paramTys, finalTy) = splitBackendArrows headTy
+              suppliedCount = length suppliedArgs
+              suppliedParamTys = take suppliedCount paramTys
+              remainingParamTys = drop suppliedCount paramTys
+              expectedPartialTy = foldr BTArrow finalTy remainingParamTys
+          if suppliedCount < length paramTys
+            && alphaEqBackendType resultTy expectedPartialTy
+            && not (null remainingParamTys)
+            && partialApplicationCanCaptureSuppliedArgs suppliedParamTys
```
soulomoon: This guard leaves higher-order supplied arguments on the old fallback path instead of packaging them into the partial closure. A checked program such as `choose : (Int -> Int) -> Int -> Int -> Int = \f \ignored \x f x` with `apply (choose inc 0)`, where `inc` is a closure-valued local, runs in the interpreter but `emit-backend` now fails with `unsupported static function argument "$g#0"`. The PR plan says the generated `BackendClosure` captures the supplied arguments, and the closure ABI can store function-typed captures as pointers, so rejecting function-typed supplied parameters here leaves accepted partial applications unsupported.

src/MLF/Backend/Convert.hs
```diff
@@ -1781,35 +1785,217 @@ convertTerm context env scope =
...
+        | (name, argTy) <- remainingParams
+        ]
+      allArgs = suppliedVars ++ remainingVars
+  (captures, body) <-
+    if isClosureHeadTerm context scope headTerm
+      then do
+        funExpr <- convertTermExpectedMode DirectLambda context env scope (Just headTy) headTerm
+        let funCapture =
+              BackendClosureCapture
+                { backendClosureCaptureName = partialFunctionCaptureName,
```
soulomoon: Capturing a closure-valued callee under the fresh `__mlfp_partial_function` name breaks closure classification for function parameters. For example, `def use : (Int -> Int -> Int) -> Int = \f apply (f 1); def main : Int = use keepLeft;` is accepted and `run-program` returns `1`, but `emit-backend` fails with `BackendClosureCallExpectedClosureValue ...`. The validator only sees the original parameter `f` through this renamed capture, so it never marks either the parameter or the capture as a definite closure before validating the generated `BackendClosureCall`.

soulomoon (commented):
Found two partial-application gaps that still reject checked source programs in the LLVM backend.

soulomoon:
Review findings require rework at commit `b75bb370953810d3dc3b8271b70596cf21c3172a`.

Findings to address:
- src/MLF/Backend/Convert.hs:1855 rejects function-typed supplied arguments for partial applications, so checked programs such as a known `choose` partial capturing a closure-valued local still fall back to static-function lowering and fail LLVM emission.
- src/MLF/Backend/Convert.hs:1903 captures closure-valued partial callees under a fresh name, which prevents closure classification for function parameters; an accepted `use f = apply (f 1)` program fails backend validation with `BackendClosureCallExpectedClosureValue`.

src/MLF/Backend/Convert.hs
```diff
@@ -1786,35 +1827,344 @@ convertTerm context env scope =
...
+  | otherwise =
+      convertTermExpectedMode DirectLambda context env scope (Just expectedArgTy) arg
+
+applicationArgumentNeedsClosureValue :: ConvertContext -> ElabTerm -> Bool
+applicationArgumentNeedsClosureValue context fun
+  | partialApplicationMentionsEvidence fun = False
+  | otherwise =
+      case stripClosureHeadTypeInsts headTerm of
+        EVar name ->
+          Set.member suppliedCount (Map.findWithDefault Set.empty name (ccClosureValueArguments context))
```
soulomoon: This demand map is only populated for top-level `CheckedBinding`s, so local let-bound higher-order helpers are missed. For example, `let use : (Int -> Int -> Int) -> Int = \f apply (f 1) in use keepLeft` is accepted by the interpreter, but the generated LLVM stores `ptr @"Main__keepLeft"` into the partial closure environment and the closure entry later treats that field as a closure record pointer before `BackendClosureCall`. That can dereference code as closure layout at runtime. Please extend the closure-value argument demand analysis to local function bindings, or otherwise package direct function arguments as closure values when a local callee underapplies them.

soulomoon (commented):
Found one remaining local-helper gap in the partial-application closure packaging path.

src/MLF/Backend/Convert.hs
```diff
@@ -1270,17 +1275,52 @@ buildConvertContext checked = do
...
+
+checkedBindingClosureValueArguments :: ConvertContext -> CheckedModule -> CheckedBinding -> Either BackendConversionError [Int]
+checkedBindingClosureValueArguments context checkedModule binding = do
+  canonicalElabTyOpen <- checkedBindingCanonicalTypeOpen context checkedModule binding
+  rawBindingTy <- convertElabType canonicalElabTyOpen
+  let bindingTy = canonicalizeBackendType context rawBindingTy
+      (_, valueTy) = splitBackendForalls bindingTy
+      (paramTys, _) = splitBackendArrows valueTy
+      (params, body) = collectLeadingLams (length paramTys) (checkedBindingTerm binding)
+  pure
```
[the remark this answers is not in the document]
soulomoon: Still applies at fd6df38: a top-level alias such as `def useAlias : (Int -> Int -> Int) -> Int = use` still skips the closure-value argument demand from `use`; `useAlias keepLeft` emits LLVM that stores `ptr @"Main__keepLeft"` and later dereferences it as closure layout.

src/MLF/Backend/Convert.hs
```diff
@@ -1786,35 +1827,344 @@ convertTerm context env scope =
...
+                        { backendClosureCaptureName = captureName,
+                          backendClosureCaptureType = headTy,
+                          backendClosureCaptureExpr = headExpr
+                        }
+                    funVar = BackendVar headTy captureName
+                bodyExpr <- liftEitherConvert (applyPartialDirectArguments funVar headTy allArgs)
+                pure (funCapture : suppliedCaptures, bodyExpr)
+              Nothing -> do
+                bodyExpr <- liftEitherConvert (applyPartialDirectArguments headExpr headTy allArgs)
+                pure (suppliedCaptures, bodyExpr)
```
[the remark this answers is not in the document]
soulomoon: Still applies at fd6df38: a checked source example like `let base : Int = 1 in apply (((\z make z) base) 2)` runs in the interpreter but `emit-backend` fails with a lambda arity mismatch through this non-variable partial-callee path.

src/MLF/Backend/Convert.hs
soulomoon: Still applies at fd6df38: local let-bound higher-order helpers are not recorded in `ccClosureValueArguments`; `let use : (Int -> Int -> Int) -> Int = \f apply (f 1) in use keepLeft` emits LLVM that stores `ptr @"Main__keepLeft"` into a partial closure environment and later loads it as a closure record.

soulomoon:
Review findings require rework at commit `fd6df38da4586d477f270066d142fc3d3363b89e`.

Findings to address:
- Unresolved review thread: Still applies at fd6df38: a top-level alias such as `def useAlias : (Int -> Int -> Int) -> Int = use` still skips the closure-value argument demand from `use`; `useAlias keepLeft` emits LLVM that stores `ptr @"Main__keepLeft"` and later dereferences it as closure layout.
- Unresolved review thread: Still applies at fd6df38: a checked source example like `let base : Int = 1 in apply (((\z make z) base) 2)` runs in the interpreter but `emit-backend` fails with a lambda arity mismatch through this non-variable partial-callee path.
- Unresolved review thread: Still applies at fd6df38: local let-bound higher-order helpers are not recorded in `ccClosureValueArguments`; `let use : (Int -> Int -> Int) -> Int = \f apply (f 1) in use keepLeft` emits LLVM that stores `ptr @"Main__keepLeft"` into a partial closure environment and later loads it as a closure record.

src/MLF/Backend/Convert.hs
```diff
@@ -1786,35 +1884,385 @@ convertTerm context env scope =
...
+              let (paramTys, finalTy) = splitBackendArrows headTy
+                  suppliedCount = length suppliedArgs
+                  suppliedParamTys = take suppliedCount paramTys
+                  remainingParamTys = drop suppliedCount paramTys
+                  expectedPartialTy = foldr BTArrow finalTy remainingParamTys
+              if suppliedCount < length paramTys
+                && alphaEqBackendType resultTy expectedPartialTy
+                && not (null remainingParamTys)
+                then do
+                  suppliedArgExprs <- zipWithM (convertTermExpectedMode DirectLambda context env scope . Just) suppliedParamTys suppliedArgs
```
soulomoon: This conversion still ignores the callee's closure-value argument demand when building captures for a partial application. If the supplied function is a direct known function rather than an already closure-valued local, `partialApplicationCanCaptureSuppliedArgs` rejects it and the old static-function path is used. For example, `choose : (Int -> Int -> Int) -> Int -> Int -> Int = \f \ignored \x f x ignored` with `apply (choose keepLeft 0)` is accepted by `run-program` but `emit-backend` fails with `unsupported static function argument "$f#0"`. The supplied `keepLeft` needs the same `convertClosureValueArgument` wrapping used for closure-demanded call arguments before packaging the partial closure.

soulomoon (commented):
Found one remaining higher-order partial-application gap.

soulomoon:
Review findings require rework at commit `7da093c5b9f7afaf60a6296c1c2cd940570f7d54`.

Findings to address:
- src/MLF/Backend/Convert.hs:1957 converts supplied partial-application arguments without honoring the callee's closure-value argument demand, so a checked program like `apply (choose keepLeft 0)` still falls back to static-function lowering and `emit-backend` fails with `unsupported static function argument "$f#0"`.

soulomoon (commented):
Found one remaining partial-application lowering gap for polymorphic supplied arguments.

src/MLF/Backend/Convert.hs
```diff
@@ -1786,35 +1941,433 @@ convertTerm context env scope =
...
+
+partialApplicationCanCaptureSuppliedArgs :: ConvertContext -> ClosureScope -> [(BackendType, BackendExpr)] -> Bool
+partialApplicationCanCaptureSuppliedArgs context scope =
+  all canCapture
+  where
+    canCapture (argTy, argExpr)
+      | isClosureConvertibleFunctionType argTy =
+          backendExprIsClosureValue context scope argExpr
+      | otherwise =
+          True
```
soulomoon, line 2467: This `otherwise = True` also treats polymorphic function values as capturable supplied arguments, but the generated closure environment cannot store a `BTForall` value. A checked program such as `def id : forall a. a -> a = \x x; def usePoly : (forall a. a -> a) -> Int -> Bool = \poly \ignored let keepInt = poly 1 in poly true; def apply : (Int -> Bool) -> Bool = \f f 0; def main : Bool = apply (usePoly id);` runs in the interpreter (`true`), but `emit-backend` fails at the generated partial closure capture with `escaping polymorphic binding "Main__id"`. Since the PR plan/issue call for accepted partial applications to lower, polymorphic supplied function values need a supported wrapper/specialization path here instead of being accepted as ordinary closure captures.

soulomoon:
Review findings require rework at commit `519c11fc4eb908f653da9a74a98c6c1e031a8b48`.

Findings to address:
- src/MLF/Backend/Convert.hs:2370 treats polymorphic supplied arguments as capturable partial-closure values; an accepted `apply (usePoly id)` program runs in the interpreter but `emit-backend` fails with `escaping polymorphic binding "Main__id"`, so the PR still does not satisfy the accepted-source partial application lowering requirement.

soulomoon (commented):
Found one remaining higher-rank partial-application lowering gap.

src/MLF/Backend/Convert.hs
```diff
@@ -1786,35 +1941,443 @@ convertTerm context env scope =
...
+partialApplicationMentionsEvidence :: ElabTerm -> Bool
+partialApplicationMentionsEvidence term =
+  any isEvidenceCaptureName (Set.toList (freeTermVariables term))
+
+partialApplicationCanCaptureSuppliedArgs :: ConvertContext -> ClosureScope -> [(BackendType, BackendExpr)] -> Bool
+partialApplicationCanCaptureSuppliedArgs context scope =
+  all canCapture
+  where
+    canCapture (argTy, argExpr)
+      | isClosureConvertibleFunctionType argTy =
```
soulomoon, line 2461: This still accepts any arrow-typed supplied argument as a closure capture, including higher-rank function types whose closure entry parameters cannot be lowered. For example, `idScore : (forall a. a -> a) -> Int`, `useHigher : ((forall a. a -> a) -> Int) -> Int -> Int`, and `main = apply (useHigher idScore)` is accepted and `run-program` returns `1`, but `emit-backend` fails in the generated partial closure with `escaping function value "__mlfp_partial_capture0"`. These non-first-order function arguments need to stay on the existing static lowering path, or the closure ABI/lowering needs to support their parameter representation, before they are treated as capturable partial-closure fields.

soulomoon:
Review findings require rework at commit `e81fca8122f61275611262a13fd16a8508221035`.

Findings to address:
- src/MLF/Backend/Convert.hs:2367 still closure-captures higher-rank supplied function arguments such as `(forall a. a -> a) -> Int`; an accepted `apply (useHigher idScore)` repro runs in the interpreter but `emit-backend` fails in the generated partial closure with `escaping function value "__mlfp_partial_capture0"`, so non-first-order function arguments still need static lowering or supported closure ABI representation.

src/MLF/Backend/Convert.hs
```diff
@@ -1786,35 +1941,508 @@ convertTerm context env scope =
...
+  BackendType ->
+  ConvertM (Maybe BackendExpr)
+convertPartialApplication context env scope term resultTy =
+  case collectApps term of
+    (rawHeadTerm, rawSuppliedArgs) ->
+      case normalizePartialApplicationSpine rawHeadTerm rawSuppliedArgs of
+        (headTerm, suppliedArgs)
+          | not (null suppliedArgs),
+            not (isConstructorHeadTerm context headTerm),
+            not (partialApplicationMentionsEvidence term) -> do
```
soulomoon: This evidence guard still skips packaging for accepted constrained partial applications. For example, with `class Pick a { pick : a -> a -> a; }`, `instance Pick Int { pick = \x \y x; }`, `keep : Pick Int => Int -> Int -> Int = \x \y pick x y`, and `main = apply (keep 1)`, `run-program` returns `1` but `emit-backend` fails with `unsupported static function argument "$f#0"`. The hidden evidence should be treated as an already supplied/non-captured argument while the remaining value underapplication is still lowered to a closure.

soulomoon (commented):
Found one constrained partial-application lowering gap.

soulomoon:
Review findings require rework at commit `945a341907fe0020332ed148a749ed650a7dfdb1`.

Findings to address:
- src/MLF/Backend/Convert.hs:2003 still skips partial-application packaging whenever hidden evidence is mentioned; an accepted constrained helper partial such as `apply (keep 1)` runs in the interpreter but `emit-backend` fails with `unsupported static function argument "$f#0"`.

src/MLF/Backend/Convert.hs
```diff
@@ -1270,18 +1302,139 @@ buildConvertContext checked = do
...
+    (_, valueTy) = splitBackendForalls bindingTy
+    (paramTys, _) = splitBackendArrows valueTy
+    (params, body) = collectLeadingLams (length paramTys) term
+
+aliasedClosureValueArguments :: ConvertContext -> ClosureScope -> BackendType -> ElabTerm -> Set.Set Int
+aliasedClosureValueArguments context scope bindingTy term =
+  case stripClosureHeadTypeInsts headTerm of
+    EVar name ->
+      Set.fromList
+        [ demandedIndex - suppliedCount
```
soulomoon: `aliasedClosureValueArguments` is returning an index relative to the alias body’s exposed parameters, but the result is stored as an index in the binding’s full parameter list. For an eta-expanded alias with leading parameters, this shifts the demanded closure argument left. For example, `use : (Int -> Int -> Int) -> Int = \f apply (f 1)` and `useAfter : Int -> (Int -> Int -> Int) -> Int = \n use` makes the demand land on index 0 instead of index 1; `useAfter 0 keepLeft` then emits LLVM that stores `ptr @"Main__keepLeft"` in the partial closure environment and the native executable segfaults when the closure call treats that code pointer as a closure record. Please offset the propagated index by the number of leading `params` already collected from this binding.

soulomoon (commented):
Found one additional partial-application closure-demand gap.

soulomoon:
Review findings require rework at commit `1637ba71ce12f476799232a437ab36c94591e26a`.

Findings to address:
- src/MLF/Backend/Convert.hs:1413 propagates closure-demand indices from aliased bodies without offsetting for leading lambda parameters already consumed by the alias binding. An accepted eta-expanded alias such as `useAfter : Int -> (Int -> Int -> Int) -> Int = \n use` records the demand on index 0 instead of index 1, so `useAfter 0 keepLeft` emits LLVM that stores `ptr @"Main__keepLeft"` into a closure environment and the native executable segfaults when that code pointer is used as a closure record.

src/MLF/Backend/Convert.hs
```diff
@@ -1270,18 +1302,141 @@ buildConvertContext checked = do
...
+  pure (canonicalizeBackendType context rawBindingTy)
+
+bindingClosureValueArguments :: ConvertContext -> ClosureScope -> BackendType -> ElabTerm -> Set.Set Int
+bindingClosureValueArguments context scope bindingTy term =
+  directClosureValueArguments bindingTy term
+    `Set.union` aliasedClosureValueArguments context scope bindingTy term
+
+bindingEvidenceValueArguments :: BackendType -> ElabTerm -> Set.Set Int
+bindingEvidenceValueArguments bindingTy term =
+  Set.fromList
```
soulomoon, line 1425: `bindingEvidenceValueArguments` only recognizes evidence parameters that appear as syntactic leading lambdas. A constrained alias can have the evidence in its declared backend type without introducing a local evidence binder, so partial packaging no longer treats the hidden evidence as already supplied. For example, adding `def keepAlias : Pick Int => Int -> Int -> Int = keep;` and then `def main : Int = apply (keepAlias 1);` is accepted and `run-program` returns `1`, but `emit-backend` fails with `Backend LLVM arity mismatch for Main__keepAlias: expected 1, got 3`. Please propagate evidence-argument indices through aliases (or derive them from the binding type/constraint metadata) the same way closure-demand indices are propagated, so constrained helper aliases still lower as accepted partial applications.

soulomoon (commented):
Found one remaining constrained partial-application gap.

soulomoon:
Review findings require rework at commit `134b5a7d96083cba5adf0422c3179389472fc0fa`.

Findings to address:
- src/MLF/Backend/Convert.hs:1384 only records evidence argument indices from syntactic leading lambdas, so constrained aliases with hidden evidence in the declared type are not recognized. An accepted `keepAlias : Pick Int => Int -> Int -> Int = keep` partial (`apply (keepAlias 1)`) runs in the interpreter but `emit-backend` fails with `Backend LLVM arity mismatch for Main__keepAlias: expected 1, got 3`.

soulomoon (commented):
Clean review. The implementation matches the PR plan and linked issue acceptance criteria for partial-application closure packaging; targeted partial-application validation passed.

Reviewed commit: `06e17ef6eaf50f4aa9c50a390b8d599f577db242`

